PR #11, opened 2018-07-12
Initial readme documenting most features

tgoodlet:
There is a slew of tests to match and verify everything documented thus far. 
Hopefully it's a decent little start :)

pinging @vodik, @dnephin, @nonsleepr and @Konstantine00 since you all expressed interest (or gave me permission to bug you) to take a look at this.

Please feel free to be honest and grill what you don't like :+1: 
Rendered version is [here](https://github.com/tgoodlet/tractor/tree/draft_readme).

Thanks!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -267,12 +267,14 @@ actor and print the results to your screen:
     import tractor
 
 
+    # this is the first 2 actors, streamer_1 and streamer_2
     async def stream_data(seed):
         for i in range(seed):
             yield i
             await trio.sleep(0)  # trigger scheduler
 
 
+    # this is the third actor; the aggregator
     async def aggregate(seed):
         """Ensure that the two streams we receive match but only stream
         a single set of values to the parent.
@@ -325,6 +327,7 @@ actor and print the results to your screen:
         print("AGGREGATOR COMPLETE!")
 
 
+    # this is the main actor and *arbiter*
     async def main():
         # a nursery which spawns "actors"
         async with tractor.open_nursery() as nursery:
--- a/README.rst
+++ b/README.rst
@@ -1,35 +1,12 @@
 tractor
 =======
-A minimalist `actor model`_ built on multiprocessing_ and trio_.
+An async-native `actor model`_ built on trio_ and multiprocessing_.
+
 
 |travis|
 
 .. |travis| image:: https://img.shields.io/travis/tgoodlet/tractor/master.svg
     :target: https://travis-ci.org/tgoodlet/tractor
-
-``tractor`` is an attempt to take trionic_ concurrency concepts and apply
-them to distributed-multicore Python.
-
-``tractor`` lets you run and spawn Python *actors*: separate processes which are internally
-running a ``trio`` scheduler and task tree (also known as an `async sandwich`_).
-
-Actors communicate with each other by sending *messages* over channels_, but the details of this
-in ``tractor`` is by default hidden and *actors* can instead easily invoke remote asynchronous
-functions using *portals*.
-
-``tractor``'s tenets non-comprehensively include:
-
-- no spawning of processes *willy-nilly*; causality_ is paramount!
-- `shared nothing architecture`_
-- remote errors `always propagate`_ back to the caller
-- verbatim support for ``trio``'s cancellation_ system
-- no use of *proxy* objects to wrap RPC calls
-- an immersive debugging experience
-- be simple, be small
-
-.. warning:: ``tractor`` is in alpha-alpha and is expected to change rapidly!
-    Expect nothing to be set in stone and your ideas about where it should go
-    to be greatly appreciated!
 
 .. _actor model: https://en.wikipedia.org/wiki/Actor_model
 .. _trio: https://github.com/python-trio/trio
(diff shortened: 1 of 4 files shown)


tractor/__init__.py
```diff
@@ -77,21 +77,33 @@ class RemoteActorError(ActorFailure):
...
-                    await chan.send({'yield': item, 'cid': cid})
-
-                log.debug(f"Finished iterating {coro}")
-                # TODO: we should really support a proper
-                # `StopAsyncIteration` system here for returning a final
-                # value if desired
-                await chan.send({'stop': None, 'cid': cid})
+                with trio.open_cancel_scope() as cs:
+                    async for item in coro:
+                        # TODO: can we send values back in here?
```
tgoodlet: @vodik interestingly enough I had a bug (just below) in this section but this comment is relevant to some of the stuff we talked about. My main outstanding concern here is that I'm not sure how to non-blocking read from the underlying socket if nothing has been `asend()` from the other side (assuming we don't send an actual `None` through on every iteration). I know you said maybe that's how it **should** work but it still seems like unnecessary overhead to me..

tgoodlet:
@vodik, @dnephin, @nonsleepr gonna merge this soon if no-one is going to comment - no big deal if you never found the time.

README.rst
```diff
@@ -0,0 +1,591 @@
...
+
+.. _nursery: https://trio.readthedocs.io/en/latest/reference-core.html#nurseries-and-spawning
+.. _causal: https://vorpus.org/blog/some-thoughts-on-asynchronous-api-design-in-a-post-asyncawait-world/#causality
+.. _cancelled: https://trio.readthedocs.io/en/latest/reference-core.html#child-tasks-and-cancellation
+
+
+Transparent function calling using *portals*
+--------------------------------------------
+``tractor`` introdces the concept of a *portal* which is an API
+borrowed_ from ``trio``. A portal may seems similar to the idea of
```
tgoodlet: typo, should be `A portal may seem similar`

README.rst
```diff
@@ -0,0 +1,591 @@
...
+
+``tractor`` is an attempt to take trionic_ concurrency concepts and apply
+them to distributed-multicore Python.
+
+``tractor`` lets you run and spawn Python *actors*: separate processes which are internally
+running a ``trio`` scheduler and task tree (also known as an `async sandwich`_).
+
+Actors communicate with each other by sending *messages* over channels_, but the details of this
+in ``tractor`` is by default hidden and *actors* can instead easily invoke remote asynchronous
+functions using *portals*.
```
vodik: I mean, this is a semantic game, but arguable function calls in python *are* messages when you consider how attribute access works in Python. Its not really all that hidden.

You could clean this up by saying "sending messages over channels to easily invoke remote ...."

README.rst
```diff
@@ -0,0 +1,591 @@
...
+
+``tractor``'s tenets non-comprehensively include:
+
+- no spawning of processes *willy-nilly*; causality_ is paramount!
+- `shared nothing architecture`_
+- remote errors `always propagate`_ back to the caller
+- verbatim support for ``trio``'s cancellation_ system
+- no use of *proxy* objects to wrap RPC calls
+- an immersive debugging experience
+- be simple, be small
```
vodik: Don't bother with this last point, its a sematics game. Small or large are *preceptions* based on perceived feature breadth and depth and fitness to the problem at hand.

I think its more interesting to instead really define your scope and list antigoals. These are things that help keep focus and a well focused code base will never feel large and boated.

README.rst
```diff
@@ -0,0 +1,591 @@
...
+
+.. warning:: ``tractor`` is in alpha-alpha and is expected to change rapidly!
+    Expect nothing to be set in stone and your ideas about where it should go
+    to be greatly appreciated!
```
vodik: This would read punchier if you put a period between "stone" and "your".

README.rst
```diff
@@ -0,0 +1,591 @@
...
+No PyPi release yet!
+
+::
+
+    pip install git+git://github.com/tgoodlet/tractor.git
+
+
+What's this? Spawning event loops in subprocesses?
+--------------------------------------------------
+Close, but not quite.
```
vodik: I like what this header is trying to do, and I think it really works if it came immediately after the main title, but I think it falls flat since you've rambled long enough about other topics.

Might want to put a transition before this title to get back to the topic of design, or just rethink the phrasing.

I don't have any good suggestions, just that its feels like its trying to be lighthearted but came off overwhelmingly.

README.rst
```diff
@@ -0,0 +1,591 @@
...
+
+The first step to grok ``tractor`` is to get the basics of ``trio``
+down. A great place to start is the `trio docs`_ and this `blog post`_
+by njsmith_.
+
+``tractor`` takes much inspiration from pulsar_ and execnet_ but attempts to be much more
+minimal, focus on sophistication of the lower level distributed architecture,
+and of course does **not** use ``asyncio``, hence **no** event loops.
```
vodik: Rewrite this.

- This is a word salad (did you forget something): "... but attempts to be much more minimal, focus on sophistication of the ..."
- I'd suggest using "focus" instead of "minimal" - as minimal is subjective and "sophistication of the lower level distributed architecture" does not invoke ideas of minimalism - the opposite, in fact (not that there's anything wrong with **not** being minimal)
- Different idea, use a different sentence: "and of course does **not** use ``asyncio``, hence **no** event loops"

README.rst
```diff
@@ -0,0 +1,591 @@
...
+.. _trio docs: https://trio.readthedocs.io/en/latest/
+.. _pulsar: http://quantmind.github.io/pulsar/design.html
+.. _execnet: https://codespeak.net/execnet/
+.. _blog post: https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/
+.. _njsmith: https://github.com/njsmith/
+
+
+A trynamic first scene
+----------------------
+As a first example let's spawn a couple *actors* and have them run their lines:
```
vodik: I think there should be a comma after "example", but I think you could just drop it:

"Let's spawn a couple *actors* and have them run through their lines:"

Maybe through in a joke about directing a movie :wink:

README.rst
```diff
@@ -0,0 +1,591 @@
...
+            )
+            gretchen = await n.start_actor(
+                'gretchen',
+                main=partial(say_hello, 'donny'),
+                rpc_module_paths=[_this_module],
+            )
+            print(await gretchen.result())
+            print(await donny.result())
+            await donny.cancel_actor()
+            print("CUTTTT CUUTT CUT!!?! Donny!! You're supposed to say...")
```
vodik: Might not want a question mark here. Why would the director be unsure if they need to call cut?

README.rst
```diff
@@ -0,0 +1,591 @@
...
+
+
+    tractor.run(main)
+
+
+Here, we've spawned two actors, *donny* and *gretchen* in separate
```
vodik: No comma after here. But maybe rewording is better: "We've just spawned two actors..."

README.rst
```diff
@@ -0,0 +1,591 @@
...
+processes. Each starts up and begins executing their *main task*
```
vodik: Careful with tense: You just wend from past (we "spawned") to present ("starts").

Probably best to stay in past tense here. "Each process executed their *main task* ...."

README.rst
```diff
@@ -0,0 +1,591 @@
...
+defined by an async function, ``say_hello()``.  The function instructs
+each actor to find their partner and say hello by calling their
+partner's ``hi()`` function using a something called a *portal*. Each
+actor receives a response and relays that back to the parent actor (in
+this case our "director").
+
+To gain more insight as to how ``tractor`` accomplishes all this please
+read on!
```
vodik: Just drop this.

README.rst
```diff
@@ -0,0 +1,591 @@
...
+
+
+Actor spawning and causality
+----------------------------
+``tractor`` tries to take ``trio``'s concept of causal task lifetimes
+to multi-process land. Accordingly ``tractor``'s actor nursery behaves
```
vodik: Comma after accordingly

README.rst
```diff
@@ -0,0 +1,591 @@
...
+similar to the nursery_ in ``trio``. That is, an ``ActorNursery``
```
vodik: to trio's nursery.

README.rst
```diff
@@ -0,0 +1,591 @@
...
+created with ``tractor.open_nursery()`` waits on spawned sub-actors to
```
vodik: Invert this sentence: "tractor.open_nursery creates ActorNursery". It moves to active voice rather than passive voice, making the chain of actions clearer.

You've also just introduced that you're talking about the trio nurseries here, so also don't blindly jump to ActorNursery - transition to it instead.

README.rst
```diff
@@ -0,0 +1,591 @@
...
+
+Notice the ``portal`` instance returned from ``nursery.start_actor()``,
+we'll get to that shortly.
+
+Spawned actor lifetimes can be configured in one of two ways:
+
+- the actor terminates when its *main* task completes (the default if
+  the ``main`` kwarg is provided)
+- the actor can be told to ``outlive_main=True`` and thus act like an RPC
+  daemon where it runs indefinitely until cancelled
```
vodik: Okay, this is something to talk about in person, but from this reading, I see dubious value to `outlive_main`.

vodik: Or, you're just assuming the reader will understand why its useful.

You've just brought it up casually while talking about benefit cancel scopes when it seems to subvert it. Its also then even weirder that you lead with that particular example as well.

README.rst
tgoodlet: Agreed, I wanted to change this sentence anyway.

README.rst
tgoodlet: Swap with *first class `pypy3.6` support* (once it comes out of course)?

README.rst
tgoodlet: See #25 and #24!

tgoodlet:
@vodik changes added!

Issue #20, opened 2018-07-25
Patch forkserver manual management into cpython stdlib


tgoodlet:
The solution to #6 required overriding some code in [multiprocessing.forkserver.py](https://github.com/python/cpython/blob/master/Lib/multiprocessing/forkserver.py) (NB the overridden code was taken from 3.8 and includes [recent changes](https://github.com/python/cpython/commits/master/Lib/multiprocessing/forkserver.py)) but should likely be patched back into the stdlib for others who might find it useful.

The minor differences introduced in `tractor`'s custom version are:
- avoiding [implicit calls](https://github.com/tgoodlet/tractor/blob/forkserver_singleton/tractor/_forkserver_hackzorz.py#L39) to `ensure_running()`
- handling a [disconnected Unix socket](https://github.com/tgoodlet/tractor/blob/forkserver_singleton/tractor/_forkserver_hackzorz.py#L53) when spawning
- handling the same disconnected socket [server side](https://github.com/tgoodlet/tractor/blob/forkserver_singleton/tractor/_forkserver_hackzorz.py#L175)

I personally think the stdlib should allow a user to avoid the implicit `ensure_running()` calls via some configuration mechanism - it doesn't seem to add to any particularly sophisticated `forkserver` resiliency and likely is apt to hide spawning problems instead of failing loudly. It also avoids creating (arguably unnecessary) auxiliary processes per sub-sub-process.

Issue #6, opened 2018-07-07
Stop forking forkservers


tgoodlet:
I haven't dug into it yet but it seems using the [forkserver method](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) causes a new server to be created for every new process created from a sub-process. Ideally we can have only one server per `tractor` *program* and have each new (sub-sub-sub..etc.) process be spawned by the top level server.

PR #21, opened 2018-07-25
Forkserver singleton

tgoodlet:
Temporary solution for #6 by overriding some stdlib code in `multiprocessing.forkserver.py`.
A proper patch should be made eventually to get this more granular server management into the stdlib. For more detailed discussion see #20.

@vodik ping ping!

### DIFF
--- a/tractor/_forkserver_hackzorz.py
+++ b/tractor/_forkserver_hackzorz.py
@@ -26,7 +26,7 @@ from multiprocessing.context import reduction
 SIGNED_STRUCT = struct.Struct('q')     # large enough for pid_t
 
 
-class AdultForkServer(ForkServer):
+class PatchedForkServer(ForkServer):
 
     def connect_to_new_process(self, fds):
         '''Request forkserver to create a child process.
@@ -55,7 +55,7 @@ class AdultForkServer(ForkServer):
                         reduction.sendfds(client, allfds)
                         break
                     except OSError as err:
-                        if err.args[0] == 9:
+                        if err.errno == errno.EBADF:
                             print(f"Bad FD {err}")
                             client = socket.socket(socket.AF_UNIX)
                             client.connect(self._forkserver_address)
@@ -248,7 +248,7 @@ def write_signed(fd, n):
         msg = msg[nbytes:]
 
 
-class AdultSemaphoreTracker(semaphore_tracker.SemaphoreTracker):
+class PatchedSemaphoreTracker(semaphore_tracker.SemaphoreTracker):
     """Stop GD ensuring everything is running...
     """
     def getfd(self):
@@ -256,8 +256,8 @@ class AdultSemaphoreTracker(semaphore_tracker.SemaphoreTracker):
         return self._fd
 
 
-_semaphore_tracker = AdultSemaphoreTracker()
-_forkserver = AdultForkServer()
+_semaphore_tracker = PatchedSemaphoreTracker()
+_forkserver = PatchedForkServer()
 
 
 def override_stdlib():
--- a/tractor/_forkserver_hackzorz.py
+++ b/tractor/_forkserver_hackzorz.py
@@ -1,0 +1,275 @@
+"""
+This is near-copy of the 3.8 stdlib's ``multiprocessing.forkserver.py``
+with some hackery to prevent any more then a single forkserver and
+semaphore tracker per ``MainProcess``.
+"""
+import os
+import socket
+import signal
+import struct
+import sys
+import errno
+import selectors
+import warnings
+
+from multiprocessing import (
+    forkserver, semaphore_tracker, spawn, process, util
+)
+from multiprocessing.forkserver import (
+        ForkServer, MAXFDS_TO_SEND, write_unsigned,
+        # _serve_one,
+)
+from multiprocessing.context import reduction
+
+
+# taken from 3.8
+SIGNED_STRUCT = struct.Struct('q')     # large enough for pid_t
+
+
+class AdultForkServer(ForkServer):
+
+    def connect_to_new_process(self, fds):
+        '''Request forkserver to create a child process.
+
+        Returns a pair of fds (status_r, data_w).  The calling process can read
+        the child process's pid and (eventually) its returncode from status_r.
+        The calling process should write to data_w the pickled preparation and
+        process data.
+        '''
+        # self.ensure_running()  # treat our users like adults expecting
+        # them to spawn the server on their own
+        if len(fds) + 4 >= MAXFDS_TO_SEND:
+            raise ValueError('too many fds')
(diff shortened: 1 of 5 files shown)


tractor/_forkserver_hackzorz.py
```diff
@@ -0,0 +1,275 @@
...
+import errno
+import selectors
+import warnings
+
+from multiprocessing import (
+    forkserver, semaphore_tracker, spawn, process, util
+)
+from multiprocessing.forkserver import (
+        ForkServer, MAXFDS_TO_SEND, write_unsigned,
+        # _serve_one,
```
tgoodlet: Can this be relied on in 3.6?

tractor/_forkserver_hackzorz.py
```diff
@@ -0,0 +1,275 @@
...
+)
+from multiprocessing.context import reduction
+
+
+# taken from 3.8
+SIGNED_STRUCT = struct.Struct('q')     # large enough for pid_t
+
+
+class AdultForkServer(ForkServer):
```
tgoodlet: Maybe `SingletonForkServer` is a better name?

tractor/_forkserver_hackzorz.py
```diff
@@ -0,0 +1,275 @@
...
+                        else:
+                            # This shouldn't happen really
+                            warnings.warn('forkserver: waitpid returned '
+                                          'unexpected pid %d' % pid)
+
+                if listener in rfds:
+                    # Incoming fork request
+                    with listener.accept()[0] as s:
+
+                        # Thing changed - be tolerant of socket disconnects
```
tgoodlet: This the 2nd change.

tractor/_forkserver_hackzorz.py
```diff
@@ -0,0 +1,275 @@
...
+            raise ValueError('too many fds')
+        with socket.socket(socket.AF_UNIX) as client:
+            client.connect(self._forkserver_address)
+            parent_r, child_w = os.pipe()
+            child_r, parent_w = os.pipe()
+            allfds = [child_r, child_w, self._forkserver_alive_fd,
+                      semaphore_tracker.getfd()]
+            allfds += fds
+
+            # This is the only part changed
```
tgoodlet: First change.

tractor/_forkserver_hackzorz.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def connect_to_new_process(self, fds):
+        '''Request forkserver to create a child process.
+
+        Returns a pair of fds (status_r, data_w).  The calling process can read
+        the child process's pid and (eventually) its returncode from status_r.
+        The calling process should write to data_w the pickled preparation and
+        process data.
+        '''
+        # self.ensure_running()  # treat our users like adults expecting
```
tgoodlet: Expect client code to call this manually, once, at server startup.

tractor/_trionics.py
```diff
@@ -26,6 +28,7 @@ def __init__(self, actor, supervisor=None):
         # self.cancel_scope = cancel_scope
         self._children = {}
         self.cancelled = False
+        self._fs = None
```
tgoodlet: Maybe `_forkserver` is better.

tractor/_forkserver_hackzorz.py
```diff
@@ -0,0 +1,284 @@
...
+        # them to spawn the server on their own
+        if len(fds) + 4 >= MAXFDS_TO_SEND:
```
vodik: As an aside, what's the fourth descriptor?

tractor/_forkserver_hackzorz.py
```diff
@@ -0,0 +1,284 @@
...
+            allfds += fds
+
+            # This is the only part changed
+            try:
+                while True:
+                    try:
+                        reduction.sendfds(client, allfds)
+                        break
+                    except OSError as err:
+                        if err.args[0] == 9:
```
vodik: Really should use `errno.EBADF` instead of 9. Also `errno` attribute maps to `args[0]` - imho clearer:

```python
import errno

...

except OSError as err:
    if err.errno = errno.EBADF
```

tractor/_forkserver_hackzorz.py
vodik: `PatchedForkServer`?

tractor/_forkserver_hackzorz.py
```diff
@@ -0,0 +1,284 @@
...
+                            print(f"Bad FD {err}")
+                            client = socket.socket(socket.AF_UNIX)
+                            client.connect(self._forkserver_address)
+                            continue
```
vodik: How many times does this hit? Might be interesting doing an os.walk on `/proc/self/fd/'

tractor/_trionics.py
```diff
@@ -53,9 +57,34 @@ def __init__(self, actor, supervisor=None):
...
+                semaphore_tracker._semaphore_tracker._fd,
+            )
+        else:
+            fs_info = (
+                fs._forkserver_address,
+                fs._forkserver_alive_fd,
+                fs._forkserver_pid,
+                semaphore_tracker._semaphore_tracker._pid,
+                semaphore_tracker._semaphore_tracker._fd,
+             ) = self._actor._forkserver_info
```
vodik: These two assignments are kinda tricky: you're building a tuple, unpacking tuples two different ways, while also chaining assignment. Its clever, but maybe borderline not readable?

At the very least, I could see someone assuming these two statements as similar - as they're written to look parallel - but they're not. The dominating block of the top constructs a tuple, while its unpacking one in the bottom. That last assignment could be visually easy to overlook.

Personal rule of thumb (imho) - the further you get to the end of a line, the less important the details should be.

I'd consider rewriting this, even if it makes it more verbose - but I don't necessarily consider this an egregiously offense either.

tgoodlet:
> but I kinda want to dig into the weird OSError/BADF error with you sometime.

@vodik definitely. We can do that as part of #20!

tractor/_forkserver_hackzorz.py
tgoodlet: iirc a couple times per proc. Yeah definitely worth further investigation.
Fwiw it seems `forkserver.py` is under pretty active development in cpython 3.8.

tractor/_forkserver_hackzorz.py
tgoodlet: Done

tractor/_forkserver_hackzorz.py
tgoodlet: No idea - ain't my code. There's 2 pipes and a socket iirc.

vodik: likely 3 pipes: stdin, stdout, stderr.

PR #25, opened 2018-08-02
Drop the "main" task via kwarg idea

### LINKED ISSUE #24: The main task idea is really really wrong!
After an in depth discussion with @vodik surrounding #11, it's so obvious now that the `main` kwarg to `ActorNursery.start_actor()` is just plain stupid and wrong. The idea of a main task is the same as starting an actor and then submitting a single function/task to that actor and then cancelling the actor once it's lone task is complete. So instead of this inverted approach (of making an actor with a *main* task the *lower level API* thus [severely complicating the internal implementation of `Actor._async_main()`](https://github.com/tgoodlet/tractor/blob/master/tractor/_actor.py#L404)) we agreed the nursery should instead get a `ActorNursery.run_in_actor()` method which internally is simply a wrapper around:

```python
async def run_in_actor(self, name: str, main: Function):
    portal = self.start_actor(name, rpc_module_paths=['my_module_with_func'])
    self._main_children.append(portal)
    return await portal.run('my_module_with_func', func_name)
    await portal.cancel_actor()
```
When the nursery block ends this actor is waited upon for it's *final result* much like is done [implicitly in the core](https://github.com/tgoodlet/tractor/blob/master/tractor/_actor.py#L212)...
Couple notes:
- the `Portal` will need to learn a *completed* state that signals the far end task is done (in the case of an async generator this happens when **either end** closes - i.e. `.aclose()`)
- the *completed state* variable will likely need to be a `trio.Event`
 
Note this also avoids the confusion in *how does the `main=my_func` module in the remote actor get figured out?* (hint it doesn't since this was using the `multiprocessing.forkserver`'s internal pickling stuff to make it happen).

~This let's us entirely drop the idea of `Portal.result()` afaik which in turn means we lose all *future*-like junk just like `trio`~ and a bunch of implementation stuff should be severely simplified!

^ Hmm not sure it should get dropped otherwise it gets tricky trying to figure out how you can call `run_in_actor()` without blocking. Maybe it's ok to look somewhat like the `Future` returned from [`loop.run_in_executor()`](https://docs.python.org/3/library/asyncio-eventloop.html#executor) as in `asyncio`? This also allows for additional `portal.run()` calls to be made to that same actor while the nursery block is open.

tgoodlet:
Stop worrying about a "main task" in each actor and instead add an
additional `ActorNursery.run_in_actor()` method which wraps calls
to create an actor and run a lone RPC task inside it. Note this
adjusts the public API of `ActorNursery.start_actor()` to drop
its `main` kwarg.

The dirty deats of making this possible:
- each spawned RPC task is now tracked with a specific cancel scope such
  that when the actor is cancelled all ongoing responders are cancelled
  before any IPC/channel machinery is closed (turns out that spawning
  new actors from `outlive_main=True` actors was probably borked before
  finally getting this working).
- make each initial RPC response be a packet which describes the
  `functype` (eg. `{'functype': 'asyncfunction'}`) allowing for async
  calls/submissions by client actors (this was required to make
  `run_in_actor()` work - `Portal._submit()` is the new async method).
- hooray we can stop faking "main task" results for daemon actors
- add better handling/raising of internal errors caught in the bowels of
  the `Actor` itself.
- drop the rpc spawning nursery; just use the `Actor._root_nursery`
- only wait on `_no_more_peers` if there are existing peer channels that
  are actually still connected.
- an `ActorNursery.__aexit__()` now implicitly waits on `Portal.result()` on close
  for each `run_in_actor()` spawned actor.
- handle cancelling partial started actors which haven't yet connected
  back to the parent

Resolves #24

This will require updates to #11 obviously.

ping @vodik!

### DIFF
--- a/tractor/_actor.py
+++ b/tractor/_actor.py
@@ -5,9 +5,9 @@ import inspect
 import importlib
 from collections import defaultdict
 from functools import partial
-from typing import Coroutine
 import traceback
 import uuid
+from itertools import chain
 
 import trio
 from async_generator import asynccontextmanager, aclosing
@@ -27,9 +27,13 @@ class ActorFailure(Exception):
     "General actor failure"
 
 
+class InternalActorError(RuntimeError):
+    "Actor primitive internals failure"
+
+
 async def _invoke(
-    cid, chan, func, kwargs,
-    treat_as_gen=False, raise_errs=False,
+    actor, cid, chan, func, kwargs,
+    treat_as_gen=False,
     task_status=trio.TASK_STATUS_IGNORED
 ):
     """Invoke local func and return results over provided channel.
@@ -47,11 +51,15 @@ async def _invoke(
             not is_async_partial and
             not is_async_gen_partial
         ):
-            await chan.send({'return': func(**kwargs), 'cid': cid})
+            await chan.send({'functype': 'function', 'cid': cid})
+            with trio.open_cancel_scope() as cs:
+                task_status.started(cs)
+                await chan.send({'return': func(**kwargs), 'cid': cid})
         else:
             coro = func(**kwargs)
 
             if inspect.isasyncgen(coro):
+                await chan.send({'functype': 'asyncgen', 'cid': cid})
                 # XXX: massive gotcha! If the containing scope
                 # is cancelled and we execute the below line,
                 # any ``ActorNursery.__aexit__()`` WON'T be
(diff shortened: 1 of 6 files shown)


tractor/_portal.py
```diff
@@ -67,34 +53,69 @@ class Portal:
...
+
+        # wait on first response msg and handle (this should be
+        # in an immediate response)
+        first_msg = await q.get()
+        functype = first_msg.get('functype')
+
+        if functype == 'function' or functype == 'asyncfunction':
+            resp_type = 'return'
+        elif functype == 'asyncgen':
+            resp_type = 'yield'
```
vodik: Have you considered using an enum here instead of strings?

tractor/_actor.py
```diff
@@ -259,63 +277,91 @@ def get_waitq(self, actorid, cid):
...
-                        continue
-                    else:
-                        ns, funcname, kwargs, actorid, cid = msg['cmd']
+        try:
+            async for msg in chan.aiter_recv():
+                if msg is None:  # terminate sentinel
+                    log.debug(
+                        f"Cancelling all tasks for {chan} from {chan.uid}")
+                    scopes = self._rpc_tasks.pop(chan, None)
+                    if scopes:
```
vodik: If you do `scopes = self._rpc_tasks.pop(chan, ())` instead, you can skip the `if`.

tgoodlet: Good point.

tractor/_portal.py
tgoodlet: Added #27 for this.

Issue #16, opened 2018-07-16
Drop `Channel.aiter_recv()`


tgoodlet:
@vodik made a point before that `Channel.aiter_recv()` is kind of pointless and we could do the same as is currently done for the [`StreamQueue`](https://github.com/tgoodlet/tractor/blob/reorg/tractor/_ipc.py#L60) such that one can simply do:

```python
for packet in chan:
   ... stuff ..
```

For one the [`__aiter__()` already lets you do this](https://github.com/tgoodlet/tractor/blob/reorg/tractor/_ipc.py#L137).
It would basically just require using the same technique where we instead define what is currently `aiter_recv()` as a private method that's called and assigned as an instance variable in `Channel.__init__()` and then `__aiter__()` just always returns that async generator instance (or is it fine as is re-calling it every time?).

PS looking at [how `go` does something similar with `range`](https://gobyexample.com/range-over-channels) might be worth thinking about here.

Issue #7, opened 2018-07-07
Arbiter.register() / unregister should use the actor uid


tgoodlet:
Since it's possible for multiple actors to use the same name use the `Actor.uid` a `tuple` - `(name, uuid1)` instead.

PR #28, opened 2018-08-06
Reg with uid and drop aiter_recv()

tgoodlet:
Resolving a couple low hanging fruit issues: #7 and #16.

### DIFF
--- a/tractor/_actor.py
+++ b/tractor/_actor.py
@@ -189,8 +189,7 @@ class Actor:
         self,
         stream: trio.SocketStream,
     ):
-        """
-        Entry point for new inbound connections to the channel server.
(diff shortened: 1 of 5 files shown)


tractor/_actor.py
```diff
@@ -523,7 +522,7 @@ def _fork_main(self, accept_addr, forkserver_info, parent_addr=None):
             if arbiter_addr is not None:
                 async with get_arbiter(*arbiter_addr) as arb_portal:
                     await arb_portal.run(
-                        'self', 'unregister_actor', name=self.name)
+                        'self', 'unregister_actor', uid=self.name)
```
tgoodlet: that should be `self.uid` no?

PR #35, opened 2018-08-20
Add type annotations to most functions

tgoodlet:
This is purely for documentation purposes for now as it should be
obvious a bunch of the signatures aren't using the correct "generics"
syntax (i.e. the use of `(str, int)` instead of `typing.Tuple[str, int])`)
in a bunch of places. We're also not using a type checker yet and besides,
`trio` doesn't really expose a lot of its internal types very well.

cc @vodik

### DIFF
--- a/tractor/_ipc.py
+++ b/tractor/_ipc.py
@@ -42,11 +42,11 @@ class StreamQueue:
                 yield packet
 
     @property
-    def laddr(self) -> (str, int):
+    def laddr(self) -> typing.Tuple[str, int]:
         return self._laddr
 
     @property
-    def raddr(self) -> (str, int):
+    def raddr(self) -> typing.Tuple[str, int]:
         return self._raddr
 
     async def put(self, data: typing.Any) -> int:
@@ -97,11 +97,11 @@ class Channel:
         return object.__repr__(self)
 
     @property
-    def laddr(self) -> (str, int):
+    def laddr(self) -> typing.Tuple[str, int]:
         return self.squeue.laddr if self.squeue else (None, None)
 
     @property
-    def raddr(self) -> (str, int):
+    def raddr(self) -> typing.Tuple[str, int]:
         return self.squeue.raddr if self.squeue else (None, None)
 
     async def connect(
--- a/tractor/_ipc.py
+++ b/tractor/_ipc.py
@@ -1,7 +1,7 @@
 """
 Inter-process comms abstractions
 """
-from typing import Coroutine, Tuple
+import typing
 
 import msgpack
 import trio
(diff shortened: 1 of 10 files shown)


vodik:
I wouldn't do it like this, imho. Stick to what mypy/typing module wants, even if its more verbose.

Using `dict` and `tuple` directly as annotations throws away useful information you could be encoding about the structure of said types: (e.g. `Tuple[int]`, `Tuple[int, str]`, `Tuple[int, ...]` and `Tuple[...]` all mean different things: tuple of a single int, tuple of a integer and string, tuple that has at least one element of type int, and tuple of any length)

But the fact that this design not only deviates away from community's approach to this problem, but is actively hostile to using `mypy` as a development aid (I run mypy as a background checker in my editor), I have zero interest in supporting it.

I'd personally reject this in any project under my control, so not interested in reviewing it.

tgoodlet:
@vodik changed it all.

I'm just gonna go ahead and try to get `mypy` going in CI in this PR. Not quite there yet but maybe I'll get it running so we can look through the problems together.

tgoodlet:
ping @vodik. Should be running `mypy` clean in CI!

tractor/_ipc.py
```diff
@@ -42,25 +43,25 @@ def __init__(self, stream):
...
         return self._raddr
 
-    async def put(self, data):
+    async def put(self, data: Any) -> int:
         async with self._send_lock:
             return await self.stream.send_all(
                 msgpack.dumps(data, use_bin_type=True))
 
-    async def get(self):
+    async def get(self) -> Any:
```
vodik: Might want to look and see if you can define a `TypeVar` for this instead: https://mypy.readthedocs.io/en/latest/generics.html

```python
T = TypeVar('T')

class StreamQueue(Generic[T]):
    async def get(self) -> Any:
        pass
```

You can get a little bit more type safety this way if you know what the return type will be situationaly:

```python
queue = StreamQueue[str](steam)
```

For a `StreamQueue` that operates over strings.

tractor/_portal.py
```diff
@@ -51,22 +57,26 @@ class Portal:
...
         self.channel = channel
         # when this is set to a tuple returned from ``_submit()`` then
         # it is expected that ``result()`` will be awaited at some point
         # during the portal's lifetime
         self._result = None
-        self._exc = None
-        self._expect_result = None
+        self._exc: Optional[RemoteActorError] = None
+        self._expect_result: Optional[
+            Tuple[str, Any, str, Dict[str, Any]]
```
vodik: You can do something like this:

```python
Result = Tuple[str, Any, str, Dict[str, Any]

self._expect_result: Optional[Result]
```

tractor/_trionics.py
```diff
@@ -99,19 +105,19 @@ def __init__(self, actor, supervisor=None):
...
-        fn,
-        bind_addr=('127.0.0.1', 0),
-        rpc_module_paths=None,
-        statespace=None,
-        loglevel=None,  # set log level per subactor
+        name: str,
+        fn: typing.Callable,
+        bind_addr: Tuple[str, int] = ('127.0.0.1', 0),
+        rpc_module_paths: List[str] = None,
+        statespace: dict = None,
```
vodik: Probably meant to fix this one?

tractor/_trionics.py
```diff
@@ -274,13 +287,13 @@ def do_hard_kill(proc):
 
 
 @asynccontextmanager
-async def open_nursery(supervisor=None):
+async def open_nursery() -> typing.AsyncGenerator[None, ActorNursery]:
```
vodik: Got the arguments backwards? Isn't the second argument the send type? I see you yielding a nursery.

tractor/_trionics.py
tgoodlet: Yup.

PR #38, opened 2018-09-08
Multi program tests

tgoodlet:
Verifies stuff still works when running the arbiter in a separate python program.
cc @vodik

### DIFF
--- a/tests/test_multi_program.py
+++ b/tests/test_multi_program.py
@@ -1,0 +1,77 @@
+"""
+Multiple python programs invoking ``tractor.run()``
+"""
+import sys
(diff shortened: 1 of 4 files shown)


tests/test_multi_program.py
```diff
@@ -0,0 +1,77 @@
...
+
+import pytest
+import tractor
+
+
+from conftest import tractor_test
+
+
+def sig_prog(proc, sig):
+    "Kill the actor-process with ctr-c."
```
tgoodlet: shit needs to be *with `sig`*.

PR #50, opened 2018-12-11
Remote task cancelling

tgoodlet:
Cancel remote async generator tasks when you call `gen.aclose()` on the local object.

Ping @vodik @Konstantine00


vodik:
Code looks good to me!

PR #52, opened 2019-01-12
Contexts and Pub-Sub

tgoodlet:
An attempt at providing a very basic implementation of a *context* similar to [that offered in `nanomsg`](https://nanomsg.github.io/nng/man/v1.1.0/nng_ctx.5).

Also adding a `@tractor.msg.pub` publisher async gen decorator system.
This allows for building inter-actor on-demand streaming APIs.

Some other bug fixes regarding remote import-time error handling is also included.

ping @vodik @Konstantine00

### DIFF
--- a/tractor/msg.py
+++ b/tractor/msg.py
@@ -2,8 +2,9 @@
 Messaging pattern APIs and helpers.
 """
 import typing
-from typing import Dict, Any
+from typing import Dict, Any, Sequence
 from functools import partial
+from async_generator import aclosing
 
 import trio
 import wrapt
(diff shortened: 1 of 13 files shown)


tractor/msg.py
```diff
@@ -0,0 +1,153 @@
...
+def pub(*, tasks=()):
+    """Publisher async generator decorator.
+
+    A publisher can be called many times from different actor's
+    remote tasks but will only spawn one internal task to deliver
+    values to all callers. Values yielded from the decorated
+    async generator are sent back to each calling task, filtered by
+    topic on the producer (server) side.
+
+    Must be called with a topic as the first arg.
```
tgoodlet: s/topic/topics sequence

tgoodlet: really need better docs here..

PR #56, opened 2019-02-16
Use trio memory channels!

tgoodlet:
`trio` deprecated the old `Queue` in `0.11.0` and introduced the new [`trio.open_memory_channel()`](https://trio.readthedocs.io/en/latest/reference-core.html#using-channels-to-pass-values-between-tasks) system. Adopt this throughout the core and adjust all tests.

In other fun news, we were also suffering from a super subtle bug in Python itself, namely:
https://bugs.python.org/issue32526

This was the root cause of a slew of issues in some UI work in `piker`.

It turns out (at least with `trio`) asyn gen functions are **not** task safe and can cause all sorts of strange issues if operated on by multiple tasks. Thanks to [in depth diagnosis on the gitter channel](https://gitter.im/python-trio/general?at=5c65bb42dc3f0523ccb6a1ac), the alternative is to introduce an async iterator derivative of `trio.abc.ReceiveChannel` which wraps an underlying receive memory channel: the deliverer of iter-actor messages to actor-local tasks.

Todo: 
- update docs examples to use memchans
- limit `trio` to `> 0.10.0`?
 

Ping @oremanj @njsmith @vodik!
Please feel free to tear this right apart.

@oremanj @njsmith the main part I would very much appreciate your eyes on is the new [async iterator wrapper](https://github.com/tgoodlet/tractor/pull/56/files#diff-bddc99a980cb82e92f51b91e449918bfR51).

### DIFF
--- a/tractor/_portal.py
+++ b/tractor/_portal.py
@@ -36,7 +36,7 @@ async def maybe_open_nursery(nursery: trio._core._run.Nursery = None):
 async def _do_handshake(
     actor: 'Actor',  # type: ignore
     chan: Channel
-)-> Any:
+) -> Any:
     await chan.send(actor.uid)
     uid: Tuple[str, str] = await chan.recv()
 
@@ -76,8 +76,11 @@ class Portal:
         await self.channel.aclose()
 
     async def _submit(
-        self, ns: str, func: str, **kwargs
-    ) -> Tuple[str, trio.Queue, str, Dict[str, Any]]:
+        self,
+        ns: str,
+        func: str,
+        kwargs,
+    ) -> Tuple[str, trio.abc.ReceiveChannel, str, Dict[str, Any]]:
         """Submit a function to be scheduled and run by actor, return the
         associated caller id, response queue, response type str,
         first message packet as a tuple.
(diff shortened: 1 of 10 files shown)


tractor/_portal.py
```diff
@@ -48,6 +48,87 @@
     return uid
 
 
+class StreamReceiveChannel(trio.abc.ReceiveChannel):
+    """A wrapper around a ``trio.abc.ReceiveChannel`` with
```
tgoodlet: shoot that would be `trio._channel.ReceiveMemoryChannel`..

tractor/_portal.py
```diff
@@ -48,6 +48,87 @@
...
+        except trio.ClosedResourceError:
+            # when the send is closed we assume the stream has
+            # terminated and signal this local iterator to stop
+            await self.aclose()
+            raise StopAsyncIteration
+        except trio.Cancelled:
+            await self.aclose()
+            raise
+        except KeyError:
+            # if 'stop' in msg:
```
tgoodlet: oof forgot to delete this comment..

tractor/_portal.py
```diff
@@ -48,6 +48,87 @@
...
+        try:
+            msg = await self._rx_chan.receive()
+            return msg['yield']
+        except trio.ClosedResourceError:
+            # when the send is closed we assume the stream has
+            # terminated and signal this local iterator to stop
+            await self.aclose()
+            raise StopAsyncIteration
+        except trio.Cancelled:
+            await self.aclose()
```
tgoodlet: need a comment here about how cancellation is relayed to the remote task

PR #68, opened 2019-03-10
Use "clean channel shutdown" in streaming example

### LINKED ISSUE #65: Use clean channel shutdown in streaming example
Thanks again to @parity3 for pointing this out.

Instead of using `None` sentinel to terminate the mult-task aggregator loops we should instead use the new [clean channel shutdown `async with` api](https://trio.readthedocs.io/en/latest/reference-core.html#clean-shutdown-with-channels). 

We need to change [the test](https://github.com/tgoodlet/tractor/blob/master/tests/test_streaming.py#L83) as well.

tgoodlet:
Resolves #65

@parity3 mind giving that a once over to be sure it solves your critique.
New streaming example is [here](https://github.com/tgoodlet/tractor/tree/close_mem_chans#a-full-fledged-streaming-service).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -391,30 +391,32 @@ and print the results to your screen:
 
            send_chan, recv_chan = trio.open_memory_channel(500)
 
-           async def push_to_q(portal):
-               async for value in await portal.run(
-                   __name__, 'stream_data', seed=seed
-               ):
-                   # leverage trio's built-in backpressure
-                   await send_chan.send(value)
+           async def push_to_chan(portal, send_chan):
+               async with send_chan:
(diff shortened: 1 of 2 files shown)


parity3:
I've only read about / seen discussions about the semantics of the clone'ing of the send channels but to my understanding, that's both cleaner, and the way I envisioned how things are supposed to work.

I don't have a testbed set up to fully test but I'm sure you do! Looks good to me.

tgoodlet:
@parity3 yup the CI passed which has an exact test for this example after figuring out the channel closing semantics.

Thanks for taking a look!

PR #75, opened 2019-04-24
Rename override module

goodboy:
More adult, sensible name for the forkserver override code.


goodboy:
Heh, really need to bug core `cpython` about these changes as well..

Issue #89, opened 2019-10-30
N-depth nursery error propagation tests


goodboy:
I ran into some blockades in #43 getting `n`-depth `trio.MultiError` testing going:
- `multiprocessing`'s forkserver is crapping out on me at a 2-depth tree with about 6 sub-actors per level.. (spawn seems to work tho?)
- we don't have error marshalling yet until #5 is dealt with; so iterating through `MultiError`s of `MultiError`s of .. isn't quite plausible yet

I'm leaving this as a note to finish this because we should have some more generalized / random tests that ensure this doesn't break under new supervisor implementations / strategies.

goodboy:
For whoever ends up getting back to this, [we do have 3-depth nursery tests](https://github.com/goodboy/tractor/blob/master/tests/test_cancellation.py#L304) now with the `trio` spawner.

It'd be nice to just fill out the parametrized set with some lesser breadth higher depth cases. I'm pretty confident at this point the the cancellation is *correct* and N-depth *should* work.

Issue #88, opened 2019-10-26
Nursery Behavior: an error/cancel during spawn?


goodboy:
Found some undefined behavior while working on #43.

## Background
When spawning actors quickly from a nursery it's possible to have the nursery-block exit and begin to wait on actors before all processes have finished startup. Thus, you may end up cancelling initializing actors due to an early failure in one of the earliest spawned processes.

The situation is different then in `trio` since in `tractor` processes are started asynchronously and are actually running in parallel, potentially on different CPUs. In trio `Nursery.start_soon()` is used which doesn't actually start processes but merely schedules them to be run on the next checkpoint.

## Proposal
In my mind actor spawning should be atomic: actors in a nursery should not begin to be cancelled by a supervisor until all spawned actors are (attempted to be) fully initialized.

This has a few benefits:
- local sub-processes will not be un-gracefully cancelled (i.e. a hard kill via signal) during startup and avoids strange errors where IPC connections are being establishing but it isn't clear if the target actor task(s) was/were begun
- common errors will be fully recognized by all executing actor's tasks having a chance to at least execute up to their first checkpoint thus allowing for a `MultiError` to be raises by the nursery indicating the "widespread" failure in many actors / tasks (more on how to implement this below)
- in theory spawning a remote actor (on another host) will be easier to handle since we have a common nursery-internal logic to wait for actor startup to complete 

Downsides might include:
- slower nursery failures for certain strategies (e.g. a once-cancels-all may take longer to teardown since we wait on all actors to first fully spawn)
- other stuff I'm missing?

## Solution
I *think* this should be surprisingly simple to implement:
1. when spawning an actor-task enter the task's uid into a nursery tracked collection
1. when the actor-task has fully initialized and [replies back with its `functype`](https://github.com/goodboy/tractor/blob/master/tractor/_portal.py#L158) remove the task uid from said collection
1. don't allow a nursery to begin executing the process waiting machinery in `ActorNursery.wait()` until this collection is empty (no external task should be able to spawn more actors once `ActorNursery.__aexit__()` is called right?)

goodboy:
Heh, so turns out the existing code already ensures invoked actor-tasks have started before any are waited upon. The issue discovered in #43 is actually me being silly about what to expect from `range(1)`...however there is something else totally broken to do with nested nurseries spawning...hoorey.

Regardless it's good to document this behavior since I haven't really thought about it in too much detail before nor added comments in the code about it.

[...]

Issue #87, opened 2019-10-22
Reliable stream termination can be solved by tracking streams per portal, per nursery?


goodboy:
As per #57 we need a system to help manage stream *teardown*, namely the explicit calling of [`StreamReceiveChannel.aclose()`](https://github.com/goodboy/tractor/blob/master/tractor/_portal.py#L82) on either intentional stream termination by a consumer/producer or by the containing `ActorNursery` during its *wait on all actors to complete or error* phase.

## what we've got right now:
- our tests that demonstrates what should be done by client code [is manually calling `aclose`](https://github.com/goodboy/tractor/blob/master/tests/test_streaming.py#L83)
- our [context api](https://github.com/goodboy/tractor#channels-and-contexts) (aka the *push* model) also doesn't have any explicit stream shutdown api from the producer end other then built-in [cancellation machinery](https://github.com/goodboy/tractor/blob/master/tractor/_actor.py#L109) which I'm not even sure will do the right thing since there's no `Portal` instance normally on the *called*/*producer* side of a connection
- consumers interacting with a `Portal` [should get all streams cancelled](https://github.com/goodboy/tractor/blob/master/tractor/_portal.py#L246) when the [actor is cancelled](https://github.com/goodboy/tractor/blob/master/tractor/_portal.py#L269) which should happen when [a nursery block is cancelled](https://github.com/goodboy/tractor/blob/master/tractor/_trionics.py#L286), however, if deamon actors are left running **and** are producers of streams it's theoretically possible that [a nursery waiter task could spin for at least a second](https://github.com/goodboy/tractor/blob/master/tractor/_trionics.py#L148) before terminating the stream; this could potentially be avoided by calling `Portal.aclose()` for any portal created by the current `ActorNursery`?

That last suggestion needs some thought; task-actor interaction in terms of how portals can be passed around between task versus actor nurseries is not clear to me yet.

goodboy:
Actually hold up, if a push producer (using `Context`) exits its decorated function [the client should be signalled that the stream stopped](https://github.com/goodboy/tractor/blob/master/tractor/_actor.py#L112) - which results in the [local results channel being closed](https://github.com/goodboy/tractor/blob/master/tractor/_actor.py#L309) and thus subsequently in the IPC channel being cancelled.

Issue #84, opened 2019-10-18
Who doesn't like multiprocessing?


goodboy:
I've been [hearing grumblings](https://trio.discourse.group/t/help-celery-5-to-choose-a-better-trio-async-based-alternative-of-python-multiprocessing/146/4) about distaste for [`multiprocessing`](https://docs.python.org/3.8/library/multiprocessing.html). I can't discount these opinions because my inability to get a reliable system with `multiprocessing` is one of the main reasons I started building this project.

I do think it would be interesting to start some work on alternatives for process spawning/forking.

A couple ideas I have offhand:
- use [`trio`'s new built-in process spawning](https://trio.readthedocs.io/en/latest/reference-io.html#spawning-subprocesses) as an alternative [spawn method](https://docs.python.org/3.8/library/multiprocessing.html#contexts-and-start-methods)
- implement our own "forkserver" on nix systems so that we don't have to [deal with the stdlib](https://github.com/goodboy/tractor/issues/20) quiffs I have
- consider alternative fork implementations such as @catern's [sfork](https://github.com/catern/sfork) which is wrapped for python as the [`supervise` package](https://github.com/catern/supervise)
- investigate how other projects such as [`distex`](https://github.com/erdewit/distex/blob/master/distex/pool.py#L262) and [`pulsar`](https://github.com/quantmind/pulsar/blob/master/pulsar/async/concurrency.py#L399) do it (I guess by looking inside [`asyncio.create_subprocess_exec()`](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.create_subprocess_exec).

Very open to opinions and ideas here!

catern:
sfork is fairly experimental (and I never released a version of supervise that uses it - all the released versions use os.fork). [rsyscall](https://github.com/catern/rsyscall) is the more robust form of it.

goodboy:
@catern thanks for the clarification :+1: and my bad for the misread.

goodboy:
Thanks to @ryanhiebert  for finding [`trio_run_in_process`](https://github.com/ethereum/trio-run-in-process) which already implements the
>use trio's new built-in process spawning as an alternative spawn method

approach!

PR #91, opened 2019-11-22
More thorough basic supervision tests

### LINKED ISSUE #43: Initial supervision tests
#42 brings in proper `trio.MultiError` support and more deterministic cancellation semantics.
More tests are needed to ensure this system is rock solid before moving onto adding different supervision strategies as in #22.

Some I can think of off-hand that aren't in the test suite are:
- *n* `run_in_actor()`, *x* `start_actor()`
  - local error causes all to cancel
    - *n* don't error but complete quickly
    - *n* don't error but complete slowly
    - *n* all error
  - internal error in `start_actor()` and `run_in_actor()`
    - *n* all error 
    - *n* some error
- propagation of `MultiError` up subactor nursery trees

More to come...

goodboy:
Attention all lurkers: code reviews and specifically criticisms are appreciated!

This is mostly to resolve #43 but I created some related issues (#87, #88, #89) after thinking and working through other types of failure cases we need to handle.

The main solution to get this working was adding proper `MultiError` propagation support in the core.

Interestingly writing the nested sub-actor error propagation tests actually found some serious holdups in the `multiprocessing` forkserver code. Even with my patches to the stdlib (which are pretty questionable) the server breaks pretty easily (causing hangs) when spawning any more then a couple "levels" (really lifetime scopes) of process tree. It's all good fodder for #84 and either writing our own forkserver or looking at the proper way to get the stdlib's version to play with our model. Surprisingly the *spawn* method works just fine so trying out [`trio_run_in_process`](https://github.com/ethereum/trio-run-in-process) has some more incentive; interested to see how it performs vs. the stdlib's version.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,7 +10,7 @@ An async-native "`actor model`_" built on trio_ and multiprocessing_.
 
 .. _actor model: https://en.wikipedia.org/wiki/Actor_model
 .. _trio: https://github.com/python-trio/trio
-.. _multiprocessing: https://docs.python.org/3/library/multiprocessing.html
+.. _multiprocessing: https://en.wikipedia.org/wiki/Multiprocessing
 .. _trionic: https://trio.readthedocs.io/en/latest/design.html#high-level-design-principles
 .. _async sandwich: https://trio.readthedocs.io/en/latest/tutorial.html#async-sandwich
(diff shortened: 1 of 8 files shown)


goodboy:
And there you have it. Windows doesn't like "nested sub-processes" even with the spawn method...

PR #92, opened 2019-11-23
Drop use of `trio.Event.clear()`

### LINKED ISSUE #78: Drop usage of Event.clear()?
As per much discussion in python-trio/trio#637 I'd like to take a look at replacing certain [uses of `Event.clear()`](https://github.com/goodboy/tractor/blob/master/tractor/_actor.py#L237) and maybe even just [events as locks](https://github.com/goodboy/tractor/blob/master/tractor/_trionics.py#L269) in general? 

In the former case(s) it seems to be used as a periodic event: the state when no peer actors are connected to the current actor.

I need to dig into that discussion further.

goodboy:
Just spin up new events instead; because apparently they're
so cheap :roll_eyes:.

Resolves #78

### DIFF
--- a/tests/test_cancellation.py
+++ b/tests/test_cancellation.py
@@ -261,13 +261,26 @@ async def test_some_cancels_all(num_actors_and_errs, start_method):
         pytest.fail("Should have gotten a remote assertion error?")
 
 
-async def spawn_and_error(num) -> None:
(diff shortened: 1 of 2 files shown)


goodboy:
Gah, windows CI definitely being inconsistent on those nested spawning tests..

PR #94, opened 2019-12-02
Log task and actor names where possible

### LINKED ISSUE #13: Log the trio task name
You can get it with `trio.current_actor().name` and likely have to make some kind of custom [formatter with `logging`](https://docs.python.org/3/library/logging.html#logging.Formatter).

goodboy:
Prepend the actor and task names in each log emission. This makes
debugging much more sane since you can see from which process and
running task the log message originates from!

Resolves #13


goodboy:
Heh, `mypy` strikes again.

PR #95, opened 2020-01-20
Support trio-run-in-process as process spawning backend

### LINKED ISSUE #93: Try `trio_run_in_process` (aka trip)
In starting to think about #84 I gave [trio_run_in_process](https://github.com/ethereum/trio-run-in-process) a [very cursory go](https://github.com/goodboy/tractor/commits/try_trip) over the last day and was able to port to the API pretty easily but I'm running into lots of "fun" getting it to work smoothly.

There's going to be a bunch of friction it seems, mostly due to taking a lot of handy stuff for granted with `multiprocessing`. 

A brief offhand list includes:
- no built-in [process name logging](https://docs.python.org/3.8/library/logging.html#logrecord-attributes) which somewhat relates to our need for task name logging per #13
- the target module python path doesn't seem to be exported the way it is with `multiprocessing` (not entirely sure which code is responsible for this but briefly looking at `spawn.py` makes me think it's to do with the [`get_preparation_data()`](https://github.com/python/cpython/blob/c09a9f56c08d80567454cae6f78f738a89e1ae94/Lib/multiprocessing/spawn.py#L150) stuff which gets [passed to the child over a pipe](https://github.com/python/cpython/blob/c09a9f56c08d80567454cae6f78f738a89e1ae94/Lib/multiprocessing/spawn.py#L124) when [the parent is launched](https://github.com/python/cpython/blob/c09a9f56c08d80567454cae6f78f738a89e1ae94/Lib/multiprocessing/popen_spawn_posix.py#L42)
  - right now we're failing when the [subproc imports a local module](https://github.com/goodboy/tractor/blob/master/tractor/_actor.py#L213) it fails (and unfortunately hangs as well - which seems to be another issue probably to do with not closing the async ctx mng correctly..)
  - a better solution then passing all this parent process data might be to just have the child process [import the target module from an absolute path](https://stackoverflow.com/a/67692)
- afaict `multiprocessing` [automatically pickles the process object on launch](https://github.com/python/cpython/blob/c09a9f56c08d80567454cae6f78f738a89e1ae94/Lib/multiprocessing/popen_spawn_posix.py#L47) and sends to the child the entire `tractor._actor.Actor` (due to it being the provider of the `target` method in our case and the [forking pickler](https://github.com/python/cpython/blob/c09a9f56c08d80567454cae6f78f738a89e1ae94/Lib/multiprocessing/reduction.py#L60) seeming to send over the entire process + any referred to objects recursively)
  - *trip* can also [send down pickled object via `args`](https://github.com/ethereum/trio-run-in-process/blob/master/trio_run_in_process/_child.py#L96) [using `cloudpickle`](https://github.com/ethereum/trio-run-in-process/blob/master/trio_run_in_process/_utils.py#L71) so we can likely make it work but the actor instance may have to be passed through explicitly

Anyone wanting to help work on this would be greatly appreciated as I've already got quite a few other things to get working and tested at the moment. I do however think that progress on this front will only bode well for the project since I've already run into many problems trying to scale the `multiprocessing` spawn and forkserver start methods - see #89. I do think we'll want to write our own `trio`-ized forkserver for nix stuff eventually as well.

goodboy:
Resolves #93.

For those interested in support for alternatives to `multiprocessing` this PR bring in the first new choice: [`trio-run-in-process`](https://github.com/ethereum/trio-run-in-process). I had to rejig the actor nursery machinery but I think it's overall resulted in more *trionic*, easier to understand code. 

*trip* (trio-run-in-process) seems to not suffer the ailments of the spawning in `mp` namely infinitely recursive, structured concurrent, actor pools seem to be quite reliable. The spawning is definitely slower then the forkserver method still but I think this is a decent first step toward *trio-izing* process launching entirely. I'm interested now to see how hard it would be to roll our own forkserver method using just `trio` APIs.

The test suite is likely to fail at first since I haven't integrated `trip` into the test generation system quite yet. Failure on Windows is expected here out of the box as well.

I would appreciate any code reviews or criticisms from lurkers ;)
ping @ryanhiebert


tests/test_rpc.py
```diff
@@ -60,7 +60,7 @@ def test_rpc_errors(arb_addr, to_call, testdir):
     if exposed_mods == ['tmp_mod']:
         # create an importable module with a bad import
         testdir.syspathinsert()
-        # module should cause raise a ModuleNotFoundError at import
+        # module should cause a raise of a ModuleNotFoundError at import
```
ryanhiebert: Warning, I'm pedantic. I'd have worded this as `module should raise a ModuleNotFoundError at import`. Shorter and grammatically correct.

tractor/__init__.py
```diff
@@ -99,15 +100,18 @@ def run(
     name: Optional[str] = None,
     arbiter_addr: Tuple[str, int] = (
         _default_arbiter_host, _default_arbiter_port),
-    # the `multiprocessing` start method:
+    # either the `multiprocessing` start method:
     # https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
-    start_method: str = 'forkserver',
+    # OR `trio-run-in-process` (the new default).
+    start_method: str = 'trip',
```
ryanhiebert: This comment is confusing, since it seems like its saying that `trio-run-in-process` is the new default, but it's actually the shortened version `trip` that is the default. I'd pick one or the other as the "correct" spelling in the code.

goodboy: Do you think the long or short name is better?

ryanhiebert: I like the short name better, but I think it would be best to be consistent with the package name. Maybe we'll eventually get a package with a shorter, more user friendly name. Then I'll be very happy to see us using that name instead.

goodboy: @ryanhiebert went with the explicit name as asked.

goodboy:
CI is :green_salad: so I think this might be good to land.

Any further lurker comments and quiffs are greatly appreciated :+1:

goodboy:
If no one is going to review or complain about this I might just merge it and hope for the best.
There's only a few more things I want to get done before pushing an alpha version onto pypi so speak now or never :wink:

ryanhiebert:
Fine by me. As long as you're happy and it's still running, don't let your momentum be stalled... Sometimes you're the only one looking.

tractor/_spawn.py
```diff
@@ -55,54 +84,200 @@ def is_main_process() -> bool:
...
+async def new_proc(
+    name: str,
+    actor_nursery: 'ActorNursery',  # type: ignore
+    subactor: Actor,
+    errors: Dict[Tuple[str, str], Exception],
     # passed through to actor main
     bind_addr: Tuple[str, int],
     parent_addr: Tuple[str, int],
-) -> mp.Process:
+    begin_wait_phase: trio.Event,
```
goodboy: Ahh dang do we need this?

goodboy: Nope. Was left over from an old approach.

goodboy:
Big props to @salotz for running the first example in the docs and finding it was totally borked...

 2a43079 fixes this and should probably fix issues others have seen on Windows to do with `__mp_main__` not found errors.  For now that commit introduces [*more manual* module loading](https://github.com/goodboy/tractor/commit/2a4307975d7c577bcaff4d3c628197406f0bb163#diff-b8f84f4d6c888a6f3d6c9c5816acc119R246) in sub-actors, something `multiprocessing` [was doing for us behind the scenes before](https://github.com/python/cpython/blob/master/Lib/multiprocessing/spawn.py#L240).

goodboy:
Just so it's noted, I actually instead just copied the `__main__` fixups stuff `multiprocessing` does into a new private module (see the commit message for deats). This technique seems to be the right way to make subprocesses behave in the expected way in terms of module loading. We only use it when **not** using the `mp` backend (i.e. for now when using `trio_run_in_process`).

tractor/__init__.py
```diff
@@ -99,16 +99,18 @@ def run(
     name: Optional[str] = None,
     arbiter_addr: Tuple[str, int] = (
         _default_arbiter_host, _default_arbiter_port),
-    # the `multiprocessing` start method:
+    # either the `multiprocessing` start method:
     # https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
-    start_method: str = 'forkserver',
+    # OR `trio_run_in_process` (the new default).
+    start_method: Optional[str] = None,
```
salotz: This should just default to the string for 'trip'. I think you set it to None here because that is what the `try_set_start_method` returns for trip. I don't think that's how that function should behave, so once that gets refactored this should just be a string type with whatever the string repr for that mode is.

tractor/__init__.py
```diff
@@ -99,16 +99,18 @@ def run(
...
     **kwargs,
 ) -> Any:
     """Run a trio-actor async function in process.
 
     This is tractor's main entry and the start point for any async actor.
     """
-    _spawn.try_set_start_method(start_method)
+    if start_method is not None:
```
salotz: relatedly wouldn't need this check here with above refactoring.

tractor/_actor.py
```diff
@@ -162,6 +170,14 @@ class Actor:
     _root_nursery: trio.Nursery
     _server_nursery: trio.Nursery
 
+    # marked by the process spawning backend at startup
+    # will be None for the parent most process started manually
+    # by the user (currently called the "arbiter")
+    _spawn_method: Optional[str] = None
```
salotz: ditto. Unless this has different semantics which is to just choose whichever was chosen in `run`, in which case the Optional would be correct.

.travis.yml
```diff
@@ -22,8 +22,19 @@ matrix:
...
-        - python: 3.7  # this works for Linux but is ignored on macOS or Windows
-        - python: 3.8
+        - name: "Python 3.7: multiprocessing"
+          python: 3.7  # this works for Linux but is ignored on macOS or Windows
+          env: SPAWN_BACKEND="mp"
+        - name: "Python 3.7: trio-run-in-process"
+          python: 3.7  # this works for Linux but is ignored on macOS or Windows
+          env: SPAWN_BACKEND="trio_run_in_process"
+
+        - name: "Pytron 3.8: multiprocessing"
```
salotz: typo

tractor/_spawn.py
```diff
@@ -18,34 +25,62 @@
...
+else:
+    import trio_run_in_process
 
-_ctx: mp.context.BaseContext = mp.get_context("spawn")  # type: ignore
+    async def proc_waiter(proc: mp.Process) -> None:
+        await trio.hazmat.wait_readable(proc.sentinel)
 
 
-def try_set_start_method(name: str) -> mp.context.BaseContext:
+def try_set_start_method(name: str) -> Optional[mp.context.BaseContext]:
```
salotz: ditto with above. Optional and Union return values is a code smell anyways

tractor/_actor.py
goodboy: @salotz for the parent-most (called arbiter) right now this will actually stay as `None` since there was no *spawn method* that created the top most process.

tractor/_actor.py
```diff
@@ -171,9 +187,22 @@ def __init__(
...
+        # will be passed to children
+        self._parent_main_data = _mp_fixup_main._mp_figure_out_main()
+
+        mods = {}
+        for name in rpc_module_paths or ():
+            mod = importlib.import_module(name)
+            mods[name] = _get_mod_abspath(mod)
+
+        self.rpc_module_paths = mods
         self._mods: dict = {}
```
goodboy: Should probably type annotate this.

tractor/_actor.py
```diff
@@ -225,11 +254,41 @@ def load_modules(self) -> None:
...
+                        parent_data['init_main_from_name'])
+                elif 'init_main_from_path' in parent_data:
+                    _mp_fixup_main._fixup_main_from_path(
+                        parent_data['init_main_from_path'])
+
+            for modpath, filepath in self.rpc_module_paths.items():
+                # XXX append the allowed module to the python path which
+                # should allow for relative (at least downward) imports.
+                sys.path.append(os.path.dirname(filepath))
+                # XXX leaving this in for now incase we decide to swap
```
goodboy: Maybe this isn't worth keeping in for now.
I can always dig up how to do it again if someone complains about the path mutating approach.

tractor/_spawn.py
```diff
@@ -3,7 +3,14 @@
 
 Mostly just wrapping around ``multiprocessing``.
```
goodboy: comment is wrong now..

tractor/_spawn.py
```diff
@@ -18,34 +25,62 @@
...
+    import trio_run_in_process
 
-_ctx: mp.context.BaseContext = mp.get_context("spawn")  # type: ignore
+    async def proc_waiter(proc: mp.Process) -> None:
+        await trio.hazmat.wait_readable(proc.sentinel)
 
 
-def try_set_start_method(name: str) -> mp.context.BaseContext:
+def try_set_start_method(name: str) -> Optional[mp.context.BaseContext]:
     """Attempt to set the start method for ``multiprocess.Process`` spawning.
```
goodboy: Doc is also out of date - doesn't include trip deats.

tractor/_spawn.py
```diff
@@ -55,54 +90,203 @@ def is_main_process() -> bool:
...
+            # local actor by the time we get a ref to it
+            event, chan = await actor_nursery._actor.wait_for_peer(
+                subactor.uid)
+            portal = Portal(chan)
+            actor_nursery._children[subactor.uid] = (subactor, proc, portal)
+
+            # unblock parent task
+            task_status.started(portal)
+
+            # wait for ActorNursery.wait() to be called
```
goodboy: Hmm is this comment right still?

tractor/_trionics.py
```diff
@@ -297,68 +160,89 @@ def do_hard_kill(proc):
...
         raise RuntimeError("No actor instance has been defined yet?")
 
-    # TODO: figure out supervisors from erlang
-    async with ActorNursery(actor) as nursery:
-        yield nursery
+    # XXX we use these nurseries because TRIP is doing all its stuff with
+    # an `@asynccontextmanager` which has an internal nursery *and* the
+    # task that opens a nursery **must also close it**.
+    errors: Dict[Tuple[str, str], Exception] = {}
+    async with trio.open_nursery() as da_nursery:
```
goodboy: shoudl probably comment on the difference between *da_* (daemon actor) and *ria_* (run in actor) nurseries..

tractor/_trionics.py
```diff
@@ -297,68 +160,89 @@ def do_hard_kill(proc):
...
+                    except trio.MultiError as merr:
+                        if err not in merr.exceptions:
+                            raise trio.MultiError(merr.exceptions + [err])
+                    else:
+                        raise
+
+                # last bit before first nursery block ends
+                log.debug(f"Waiting on all subactors to complete")
+                anursery._join_procs.set()
+            # ria_nursery scope
```
goodboy: should probably mark the same for the `da_nursery` scope

tractor/__init__.py
goodboy: >understand the whole multiprocessing module load thing

If you're talking about the forkserver hack stuff, yeah don't worry about it. I'm starting to think we'll drop all that to write our own anyway.

tractor/_spawn.py
goodboy: The thing is for `trip` there is no context possible to return, nor is this function called in that case.
I'm not sure how else to specify the output if it's not this?

salotz: Either make a wrapper function that does produce a context or this function simply decides on the start method given the platform yadayada and then outputs a string/enum. I don't think it blocks this PR btw

tractor/__init__.py
goodboy: @salotz so as you can see from the build triggered after fb93336 which failed, on windows we need to have a different default. If `'trio_run_in_process'` is passed to `try_set_start_method()` we're going to get an error on windows (I think this is correct since we want that function to explicitly fail when an incorrect method is chosen on the platform).

This is the small dilemma, on different systems we require different default start methods. We can either leave it as I had it -> `None` is default and implies the correct actual default for the system is chosen internally *or* we move that same logic into `tractor.run()` so that's it's more clear that's what's happening?

Interested in your opinion here :)

goodboy: Yeah and the other reason to leave that default as `None` and *not* have the *default-for-the-system* logic here would be that when we get alternatives for windows then we'll want to pass through the user chosen value and validate it.

tractor/_spawn.py
goodboy: KK we can discuss further in  #101

Issue #99, opened 2020-01-30
Sphinx docs


salotz:

goodboy:
I think for a first stab at this we just literally put up a single file `index.rst` with the contents from our current `README.rst` and then use the `..literalinclude::`s as suggested in #102.

goodboy:
Closed via #110. Not sure why I didn't link it in the commit..

Issue #98, opened 2020-01-30
doctests for examples


salotz:
Implement doctests for examples in readme. Probably should just set a proper sphinx doc subtree before this.

goodboy:
@salotz recently discovered when testing #95 that the first examples in the docs were totally broken :crying_cat_face: 

This is now fixed but we need proper doc tests that ensure the examples we write actually work.

Some options:
1. just [doctest](https://docs.python.org/3.8/library/doctest.html) (probably need this for snippets in doc strings anyway?)
1. make an `examples/` directory with modules containing example code and then run them using `pytest` and include them in sphinx docs (once we do #99) using a [`..literalinclude::` directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude)

I'm partial to 2. just because it means code as code, once, in one place, as files peeps can `wget`/`curl` and if the content changes then the docs reflect it instead of the other way around.

goodboy:
For running examples as scripts https://github.com/kvas-it/pytest-console-scripts might also be an interesting option.

salotz:
I am remembering now that in doctest you have to write things that looks like a command prompt. So it kind of mimics interactive usage. I don't think that is appropriate for the examples which run many lines, functions etc. So that script tester + include is the way to go.

PR #61, opened 2019-03-06
Spawn method support

tgoodlet:
Resolves #60 and hopefully @chrizzFTD's #59.

PR #102, opened 2020-01-31
Test docs examples as scripts

goodboy:
A per #98 we need tests for examples from the docs as they would be run
by a user copy and pasting the code. This adds a small system for loading
examples from an "examples/" directory and executing them in
a subprocess while checking the output.

I am open to other suggestions but I think this approach will work well along with plans to literal include examples in sphinx docs as mentioned in #99.

@salotz @ryanhiebert give this a peak if you don't mind :)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -108,7 +108,6 @@ the hip new film we're shooting:
 .. code:: python
 
     import tractor
-    from functools import partial
 
     _this_module = __name__
     the_line = 'Hi my name is {}'
--- a/examples/__main__.py
+++ b/examples/__main__.py
@@ -1,0 +1,11 @@
+"""
+Needed in windows.
+This needs to be the main program as it will be
+called '__mp_main__' by the multiprocessing module
+
+"""
+if __name__ == '__main__':
+    import multiprocessing
+    multiprocessing.freeze_support()
+    from . import test_example
+    test_example.tractor.run(test_example.main, start_method='spawn')
--- a/examples/__main__.py
+++ b/examples/__main__.py
@@ -1,11 +1,19 @@
 """
-Needed in windows.
-This needs to be the main program as it will be
-called '__mp_main__' by the multiprocessing module
+Needed on Windows.
+
+This module is needed as the program entry point for invocation
+with ``python -m <modulename>``. See the solution from @chrizzFTD
+here:
+
+    https://github.com/goodboy/tractor/pull/61#issuecomment-470053512
 
 """
 if __name__ == '__main__':
     import multiprocessing
     multiprocessing.freeze_support()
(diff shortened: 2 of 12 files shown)


salotz:
Thats the idea with the literal include for the main documentation, except in the reverse direction. Not sure if the README in the toplevel will be able to be included in that though without symlinks or something. I'll take a look around for some tools to "tangle" (as it is typically called in literate programming) code blocks. I do this pretty regularly in emacs org mode. Not sure what the preferred README format is for you guys though. I think there are some markdown tanglers that will do the same as well.

salotz:
I'll test the other examples and then approve. Might as well get them all in.

goodboy:
@salotz sweet. Yeah feel free to take over this branch if you feel so inclined :+1:

goodboy:
@ryanhiebert yeah the duplication can be removed once we move the docs to sphinx (#99 which needs to be done since they're already getting a bit unwieldy to be a on a single page) where we can [..literalinclude::](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) the example scripts into the docs instead of the other way around.

goodboy:
Well windows is clearly unhappy..
I wonder if someone with such a system can take a look :nerd_face:

goodboy:
Interesante. It looks as though an issue I found earlier to do with mp and the context not being setup yet is exhibiting here on windows. Gonna start by printing the whole traceback and see if that gives some insight to this racy error.

goodboy:
Huh so fixing the one problem lead to another...

It seems we're hanging on the example test as per [this build](https://travis-ci.org/goodboy/tractor/builds/647617375?utm_source=github_status&utm_medium=notification).  I'm not sure if it's the same issue @chrizzFTD found in  #61 or not? We might need someone with manual access to a Windows machine to take a look. I wonder if the solution @chrizzFTD mentioned [here](https://github.com/goodboy/tractor/pull/61#issuecomment-470053512) would solve this?

goodboy:
@ryanhiebert @salotz yeah so one of things that's going to be "annoying" is we will have some duplication between examples in `examples/` and what's in the tests regardless since the tests take some of the documented examples and parameterize them with things like different delays to cancellation or different numbers of actors spawned. I'm not sure we can really avoid the duplication really. Maybe someone will come up with a bright idea to simplify it all without adding confusion?

goodboy:
Welp, it seems @chrizzFTD's solution did indeed work so I think this is ready to land!

goodboy:
Lol I'm really surprised I got this working last night.

I just want credit the new drink my friend and I created, *el bucanero* for this work.

PR #105, opened 2020-02-07
Add `Portal.run_func()`: run local funcs remotely

### LINKED ISSUE #69: Allow passing function refs to Portal.run() ?
@parity3 had the same thought as I: why can't `Portal.run()` accept a direct function reference and then we simply lookup the functions module and name [much like we do in `ActorNursery.run_in_actor()`](https://github.com/tgoodlet/tractor/blob/master/tractor/_trionics.py#L109)? This would of course have the implicit assumption that the remote actor has the chosen function's module code loaded for use.

The only question is really how to expose both ways on `Portal`?
- use a different method, say `run_func()`?
- allow passing an explicit `func` or `ref` kwarg?
- allow the first arg to `Portal.run()` to be either a module name or a func ref and then in the named case make `name` a required kwarg?

goodboy:
This is to address the request from a couple people (@parity3, @ryanhiebert) for the ability to
not have to specify the explicit module path for a callable to be
invoked in a remote actor when the desired function is defined locally.

This makes the API to run a local func look more like
`trio.Nurser.start_soon(my_func, *args)` which is simpler and much more
"shorthand". In this case we simply introspect the function/callable
object and send that info to the remote actor to be *looked up locally*
and scheduled.

I'm not sold on making `Portal.run()` take function objects since it simply won't easily translate to multi-host setups where remote code is **not** the same as local code. If this kind of API was sane you'd likely have seen it show up in widespread app layer IPC protocols like HTTP. It makes more sense to me to have a url/explicit module lookup system to determine task endpoints/entrypoints. With this premise in mind this PR proposes that `Portal.run()` stay as is (maybe later accepting a URL-like syntax) and `Portal.run_func()` be used to easily run a locally defined function in a remote actor.

Resolves #69

Note this is partially to open discussion since I'm open to other ideas of course.
This also still needs tests!

### DIFF
--- a/tractor/_portal.py
+++ b/tractor/_portal.py
@@ -4,7 +4,7 @@ Portal api
 import importlib
 import inspect
 import typing
-from typing import Tuple, Any, Dict, Optional, Set
+from typing import Tuple, Any, Dict, Optional, Set, Callable
 from functools import partial
 from dataclasses import dataclass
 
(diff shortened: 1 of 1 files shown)


goodboy:
@ryanhiebert really appreciate this feedback! 

Maybe we can continue this discussion in #69?
I do think "trio across processes" is something I'd like to target and maybe it is as simple as making, as you said, *the registration of the functions to names in a registry should be very explicit*.

This is something that `rpyc` does with it's [new style services](https://rpyc.readthedocs.io/en/latest/tutorial/tut3.html#part-3-services-and-new-style-rpyc) and [aiomass](https://gitlab.com/sscherfke/aiomas) does with it's `@aimos.expose` api.

I'll follow up in the original issue referencing some of what you've wrote here.

goodboy:
As discussed this will be addressed in a follow up PR relating to both #69 and #163.

PR #106, opened 2020-02-07
Add script "guards" to docs examples

goodboy:
This was originally bundled in #102 but the windows CI there has blocked
that from landing quickly. These examples need to be fixed stat since
I've had at least a couple people notice it now when first trying out
the project.

Ping @salotz and @dxflores.


goodboy:
Man, `trip` keeps failing on that nested spawning test on py3.8..
I guess at least the failure seems consistent?

Issue #120, opened 2020-06-23
Running asyncio actors using "guest mode" vs. anyio task groups


goodboy:
`trio` [recently added "guest mode"](https://github.com/python-trio/trio/pull/1551/files) a very brilliant solution to integration with foreign event loops and we've used it with great success in `piker` for charting with `Qt`.

I'm wondering if it's sensible to create a new actor [entry point](https://github.com/goodboy/tractor/blob/master/tractor/_actor.py#L574) that could be run using an explicit `Nursery.start_aio_actor()`? In `piker` there is a need for something like this being built-in to avoid hoop jumping with `trio-asyncio`. Guest mode makes a lot of sense since `tractor` machinery can pretty much run unmodified and can stay written in`trio` despite being run "on top of" an `asyncio` loop.

The main comparison is whether to do this versus trying to make `anyio` "work".

### using guest mode
- i/o with `asyncio` will likely be done over queues/channels such as in the primary [guest mode example](https://github.com/python-trio/trio/blob/master/notes-to-self/aio-guest-test.py)
- this has the implication that we likely lose any ability for subtask error propagation from `asyncio`
  - note this is still a problem with `anyio` (see below)
- `uvloop` should be supported (something we don't get with `trio-asyncio`)
- streaming should mostly work out-of-the box (though we may need to cook up an api for `asyncio` specific "wrapping") 
- "remote" `asyncio` async function entry points may need to be decorated such that the correct `tractor` entry point can be configured at import time
- there may be unforeseen latency in the queue/channel translation

### adopting `anyio`
- internals need to potentially be re-written to use `anyio`'s [*task groups*](https://anyio.readthedocs.io/en/latest/tasks.html), and generally, the `anyio` compat apis
  - I'm personally not huge on this as one of the primary goals for `tractor` was to **encourage and motivate** use of `trio` (which ideally gets us things like rust loop speedups in the core :wink:)
- the implications for maintaining proper cancellation semantics may still be lost:
   >Tasks spawned by these “native” libraries on backends other than trio are not subject to the cancellation rules enforced by AnyIO
- there may be latency introduced by the compat layer?
- I'm not sure we'll ever have a need for `curio` support?
- `asyncio` may end up adopting SC concepts and apis natively at which point `anyio` may be come somewhat superfluous given that guest mode exists?

I'd appreciate input from others. I'm likely going to experiment with a guest mode shim in `piker` for now and see if I run into anything dangerous :surfer:

goodboy:
[...]

Issue #117, opened 2020-05-27
Let's roll our own subproc spawner ("the way the experts would")


goodboy:
After all the headaches from trying to integrate `trio-run-in-process`:
- #115, #96 as local standouts for which I created ethereum/trio-run-in-process#4
- the mess of 3 repos that aren't doing anything novel and overlap each other as per discussion in ethereum/trio-run-in-process#6
- the over complexity of the design and spawning logic (as discussed by me [here](https://github.com/ethereum/trio-run-in-process/issues/6#issuecomment-634447691) and confirmed by nathaniel [here](https://github.com/python-trio/trio/issues/5#issuecomment-529272338))
- the divergence of [goals of the author](https://github.com/ethereum/trio-run-in-process/issues/6#issuecomment-634727926) from [our own](https://tractor.readthedocs.io/en/latest/?badge=latest#philosophy)

I think it's time to move on to more general pastures :wink: 

There was some good refactoring and re-design that came out of this in #95 including the discovery that stdstreams if semaphored correctly can allow for `pdb++` to somewhat work out of the box (towards #113).

I would like to discuss with `anyio` and `trio` core devs about design and goals to eventually factor this functionality into the core async lib as necessary.

ryanhiebert:
Thanks for this. Yeah, I think the approach you're mentioning makes sense. And, when it's done right, I think it absolutely makes sense for it to be in the core libraries.

goodboy:
Relevant links to get this rolling:
- [`anyio` subproc support + todo](https://github.com/agronholm/anyio/issues/9)
- discussion about [worker pools in `trio`](https://github.com/python-trio/trio/issues/5)
- discussion about [spawning processes from nurseries](https://github.com/python-trio/trio/issues/1104)
- the tagged [subprocess issue list from `trio`](https://github.com/python-trio/trio/issues?q=is%3Aissue+is%3Aopen+label%3Asubprocesses+)

PR #121, opened 2020-06-29
Infect `asyncio`

goodboy:
[README blurb](https://github.com/goodboy/tractor/tree/infect_asyncio#infected-asyncio-mode) summarizing this new feature.

What I hope to be a successful attempt at solving #120.

This gets us the following very interesting functionality:
- ability to spawn an actor that has a process entry point of `asyncio.run()`
- the `asyncio` actor embeds `trio` using the [new guest mode](https://github.com/python-trio/trio/blob/master/notes-to-self/aio-guest-test.py) which then enters the `tractor.Actor._async_main()` entry point thus engaging all the std IPC machinery
- the actor is now able to accept requests for async functions which can utilize the new `tractor.to_asyncio.run()` method which allows spawning tasks in the host `asyncio` loop and waiting/streaming the result(s) back into the `trio` task
- results can then be relayed upwards to the parent actor as per normal operation

It ended up being possible to accomplish per task error propagation by simply doing a little state passing inside the callback passed to `asyncio.Task.add_done_callback()` and thus not requiring the use of `anyio`. In fact I'm not even sure it's possible to do this with the task groups offered by `anyio` due to the nature of spawning `asyncio` tasks *from* `trio` tasks and providing an api to acquire the results on demand.

The main idea here is that a new `tractor` actor can be spawned and async functions that will run under `trio` can make calls to spawn `asyncio` tasks using the new `tractor.to_asyncio.run()` routine which acts basically like `tractor.Portal.run()` but without any IPC stuff going on underneath. All IPC is done as normal once results are back in the (embedded) `trio` task.

This basically makes `tractor` integration with `asyncio` frameworks a super cinch now :)

ping @njsmith @oremanj.
If either of you have time to look at this and tell if there's a more *trionic* way it'd be grealy appreciated :smile_cat: 

Further things to be done:
- [x] tests for all of this
- [x] docs & examples
~- [ ] potentially support the existing `trip` backend using the even newer *inside-out-guest-mode* (python-trio/trio#1652) though the plan is to deprecate this anyway (#117)~ trip is being removed in #127.
- ~[ ] using the even newer *inside-out-guest-mode* (python-trio/trio#1652)?~ not yet approved in trio core.

Provisional test list:

- [x] call async func
- ~[ ] call async generator and stream from it~ no longer supported
- [x] error from async func
- [x] error during async streaming
- [x] all the above in remote actors
- ~[ ] `@stream` compatibility?~ unneeded with new api
- ~[ ] `trio_trio` / `from_trio` injection~ don't member why i wrote this
- [x] simple asyncio task error propagation
- [x] trio cancels aio on child side
- [x] cancel via `Portal.cancel_actor()`
- [x] cancel via `trio.move_on_after()` around actor nursery
- [x] asyncio cancels itself and causes a `trio` cancellation
- [x] interloop channel streaming via `to_asyncio.open_channel_from():`
- [x] trio error kills interloop channel in SC style
- [x] trio closes interloop channel early and channel exits
- [x] asyncio error propagates to trio task with interloop channel
- [x] small trio-asyncio echo server

### DIFF
--- a/tractor/to_asyncio.py
+++ b/tractor/to_asyncio.py
@@ -6,7 +6,7 @@ import inspect
 from typing import (
     Any,
     Callable,
-    AsyncGenerator,
+    AsyncIterator,
     Awaitable,
     Union,
 )
@@ -33,7 +33,7 @@ async def run_coro(
 
 async def consume_asyncgen(
     to_trio: trio.MemorySendChannel,
-    coro: AsyncGenerator,
+    coro: AsyncIterator,
 ) -> None:
     """Stream async generator results back to ``trio``.
 
@@ -50,7 +50,7 @@ async def run_task(
     qsize: int = 2**10,
     _treat_as_stream: bool = False,
     **kwargs,
-) -> Union[AsyncGenerator, Any]:
+) -> Any:
     """Run an ``asyncio`` async function or generator in a task, return
     or stream the result back to ``trio``.
     """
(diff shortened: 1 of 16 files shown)


tractor/to_asyncio.py
```diff
@@ -0,0 +1,99 @@
...
+    coro = func(**kwargs)
+
+    cancel_scope = trio.CancelScope()
+
+    # start the asyncio task we submitted from trio
+    # TODO: try out ``anyio`` asyncio based tg here
+    task = asyncio.create_task(_invoke(from_trio, to_trio, coro))
+    err = None
+
+    # XXX: I'm not sure this actually does anything...
```
goodboy, line 70: Old; it does.

tractor/to_asyncio.py
```diff
@@ -0,0 +1,99 @@
...
+    def cancel_trio(task):
+        """Cancel the calling ``trio`` task on error.
```
goodboy, line 72: and save any error.

goodboy:
Provisional test list:

- [ ] call async func
- [ ] call async generator and stream from it
- [ ] error from async func
- [ ] error during async streaming
- [ ] all the above in remote actors
- [ ] `@stream` compatibility?
- [ ] `trio_trio` / `from_trio` injection
- [ ] cancellation from a local `trio` nursery
- [ ] cancellation from a remote `tractor` nursery

guilledk (approved):

tractor/_portal.py
```diff
@@ -102,14 +105,25 @@ async def aclose(self):
...
-        if self._rx_chan._closed:
+        rx_chan = self._rx_chan
+        stats = rx_chan.statistics()
+
+        if rx_chan._closed:
             log.warning(f"{self} is already closed")
             return
 
+        if stats.open_receive_channels > 1:
+            # if we've been cloned don't kill the stream
```
richardsheridan, line 116: I think you still need to call `rx_chan.close()` here?

goodboy: @richardsheridan nice catch. 

Yeah i feel like you're right.
I wonder why I didn't notice this being a problem (yet).

tractor/_portal.py
```diff
@@ -102,14 +105,25 @@ async def aclose(self):
...
+            log.debug("there are still consumers running keeping stream alive")
+            return
+
         if self._shielded:
             log.warning(f"{self} is shielded, portal channel being kept alive")
             return
 
+        # close the local mem chan
+        rx_chan.close()
```
goodboy, line 125: So this likely needs to be duplicated for the cloned case above.

goodboy:
CI is currently held up by 2 issues:
- https://github.com/python-trio/trio-typing/issues/48
- the coming fix for #265 which will likely land with #261

goodboy:
I also don't plan to add a full `asyncio`-side memory channel API *yet* mostly because we probably need an in depth design discussion about how to both implement and present the interface.

I will be making a follow up issue for that likely this week.

goodboy:
I made https://github.com/goodboy/tractor/issues/273 to follow up re: inter-loop-task channel APIs.

This should also resolve #120 for now since I'm pretty convinced unless you use `anyio` in SC-style wholesale, you aren't getting any benefit from using it *with `asyncio`*. OTOH, wwith our infected mode style you are limited to one-to-one task relations but you are guaranteed SC since every `asyncio` task lifetime is directly linked to a `trio` task lifetime.

guilledk (approved):

PR #127, opened 2020-07-20
First attempt at removing trip & updating hazmat -> lowlevel

guilledk:
This branch started as an attempt to remove all `RuntimeWarnings` about `trio.hazmat` deprecation but also turned into dropping [ethereum/trio-run-in-process](https://github.com/ethereum/trio-run-in-process).

- Removed "trio-run-in-process" as a posible backend and added "trio".
- Added `tractor._child` which is the file that gets executed as a module when `run_in_process` is called.
- Based pickle logic on @njsmith comment in: https://github.com/python-trio/trio/issues/5#issuecomment-529272338

I'm getting a bunch of fails on tests and hang, I'm pretty sure it has to do with the tractor machinery not being setup right in the newly spawned processes.

### DIFF
--- a/tractor/_spawn.py
+++ b/tractor/_spawn.py
@@ -158,13 +158,24 @@ async def cancel_on_completion(
 
 @asynccontextmanager
 async def run_in_process(async_fn, *args, **kwargs):
+    # get arguments that are actors
+    actors_in_args = [
+        arg for arg in args if isinstance(arg, Actor)
+    ]
+    debug_args = []
+    # assume the first is the actor running the function
+    if len(actors_in_args) > 0:
+        subactor = actors_in_args[0]
+        debug_args.append(subactor.uid[0])
+        debug_args.append(subactor.uid[1])
+
     encoded_job = cloudpickle.dumps(partial(async_fn, *args, **kwargs))
     p = await trio.open_process(
         [
             sys.executable,
             "-m",
             _child.__name__
-        ],
+        ] + debug_args,
         stdin=subprocess.PIPE
     )
 
--- a/tractor/_spawn.py
+++ b/tractor/_spawn.py
@@ -168,11 +168,13 @@ async def run_in_process(async_fn, *args, **kwargs):
         stdin=subprocess.PIPE
     )
 
+    # send over func to call
     await p.stdin.send_all(encoded_job)
 
     yield p
 
-    #return cloudpickle.loads(p.stdout)
+    # wait for termination
+    await p.wait()
 
 
 async def new_proc(
--- a/tractor/_spawn.py
+++ b/tractor/_spawn.py
@@ -1,14 +1,18 @@
 """
 Machinery for actor process spawning using multiple backends.
 """
+import sys
 import inspect
+import subprocess
 import multiprocessing as mp
 import platform
 from typing import Any, Dict, Optional
+from functools import partial
 
 import trio
+import cloudpickle
 from trio_typing import TaskStatus
-from async_generator import aclosing
+from async_generator import aclosing, asynccontextmanager
 
 try:
     from multiprocessing import semaphore_tracker  # type: ignore
@@ -21,12 +25,12 @@ except ImportError:
 from multiprocessing import forkserver  # type: ignore
 from typing import Tuple
 
-from . import _forkserver_override
+from . import _forkserver_override, _child
 from ._state import current_actor
 from .log import get_logger
 from ._portal import Portal
 from ._actor import Actor, ActorFailure
-from ._entry import _mp_main, _trip_main
+from ._entry import _mp_main, _trio_main
 
 
 log = get_logger('tractor')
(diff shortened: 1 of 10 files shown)


goodboy:
@guilledk amazing :partying_face:, thanks for taking an attempt that this.

I cloned your branch and tried to run the suite with `pytest -svvv tests --pdb` and see a bunch of interesting errors:

1.
```python
Task <bound method Actor.cancel of <tractor._actor.Actor object at 0x7fc8307a6880>> was likely cancelled before it was started
  File "/home/goodboy/repos/tractor/tractor/_child.py", line 10, in <module>
    cloudpickle.dump(sys.stdout.detach(), result)
  File "/home/goodboy/repos/tractor/env/lib/python3.8/site-packages/cloudpickle/cloudpickle_fast.py", line 47, in dump
    CloudPickler(file, protocol=protocol, buffer_callback=buffer_callback).dump(obj)
  File "/home/goodboy/repos/tractor/env/lib/python3.8/site-packages/cloudpickle/cloudpickle_fast.py", line 429, in __init__
    Pickler.__init__(self, file, protocol=protocol, buffer_callback=buffer_callback)
TypeError: file must have a 'write' attribute
```

2.
```python
  File "/usr/lib64/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib64/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/goodboy/repos/tractor/tractor/_child.py", line 13, in <module>
    cloudpickle.dump(sys.stdout.detach(), err)
ValueError: underlying buffer has been detached
```

3.
```python
/usr/lib64/python3.8/runpy.py:127: RuntimeWarning: 'tractor._child' found in sys.modules after import of package 'tractor', but prior to execution of 'tractor._child'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
```

I think the last one *might* be fixed by changing [the check for the spawn method](https://github.com/goodboy/tractor/blob/master/tractor/_actor.py#L259) in `Actor.load_modules()`.

The other ones I'm note entirely sure what's going on yet; looks like stdio getting torn down too early?

goodboy:
@guilledk the other thing is you [dropped all the logging setup code](https://github.com/goodboy/tractor/pull/127/files#diff-6b6d7179b99ec7568d32065017c72fb4L65) in the child entrypoint which makes it so you can't see child logging which is **really** useful for debugging.

goodboy:
@guilledk see my change set appended to yours:
https://github.com/guilledk/tractor/compare/drop-trip-update-trio...goodboy:drop-trip-update-trio

I think this *should* make the tests run clean :surfing_woman:

goodboy:
@guilledk i made a PR to your branch so I think if you merge that and update then it should run the CI here?

goodboy:
@guilledk woot!

Looks like tests passed but `mypy` run is dying.

Check the `.travis.yml` to see how to run the mypy checks.

guilledk:
Hey so this commit is to check if the way I'm getting the actor from the passed arguments in `run_in_process` is correct.

Right now I just search the passed `*args` for the first instance of `Actor` and assume that's the right one, so I might as well use a named argument for that, but I don't know if you need this function to be generic or not.

Next commit I'll also check the passed arguments in the `_async_main`

tractor/_spawn.py
```diff
@@ -153,6 +156,38 @@ async def cancel_on_completion(
         await portal.cancel_actor()
 
 
+@asynccontextmanager
+async def run_in_process(async_fn, *args, **kwargs):
```
goodboy, line 160: To make this simpler why don't we just make the signature of `run_in_proc()` something like:

```python
async def run_in_process(subactor, async_fn, *args, **kwargs):
    debug_args = list(subactor.uid)
    ...
```

I do think it would be neat to accept these over `sys.argv` in `_child.py` and verify just as a sanity check.
It will for sure help with debugging using `pstree` (or whatever).

Heck, maybe it actually makes for sense to do this then passing the `Actor` type at all?
In theory we can avoid passing any complex types whatsoever as long as `Actor._async_main()` receives the appropriate data that `Actor.__init__()` normally does, :thinking:.

Not sure if that's overthinking it.
I wonder if that means we could avoid `cloudpickle` entirely?
We can maybe just use `msgpack` in that case?

goodboy:
@guilledk try rebasing onto the latest `infect_asyncio` *head* (`git rebase upstream infect_asyncio`, or whatever you've got this repo reffed as instead of `upstream`) to get some fixes I put up for `mypy`.

Ideally we can factor out the commonalities into your branch but I've got to look through it.

tractor/_spawn.py
goodboy: I think at the least in terms of dropping `cloudpickle` we should make an issue and tackle it later.

goodboy:
Closing in favor of the new #128.

Issue #115, opened 2020-02-14
trip doesn't do control-C right...


goodboy:
Thought I messed something up with the `trio_run_in_process` port but testing with some `piker` apps I realized that everything works fine with `ctl-c` when using `start_method='forkserver'`. 

See ethereum/trio-run-in-process#4 for the deats!

goodboy:
Heh, also seems like it's being somewhat unmaintained as I've yet to receive any feedback on this issue..

ryanhiebert:
Bummer.

goodboy:
@ryanhiebert indeed. I'm even fine to fix it myself but I'd rather do that in a fork and start working forwards if no-one is going to provide feedback..

Issue #113, opened 2020-02-11
Proper multiprocessing-native debugger wen?


goodboy:
In `tractor` every actor is both a (potential) *client* and *server*. Concocting concocting a "native" feeling "remote debugger" shouldn't be that bad (they thought naively..), right?

I did a little digging and asked all the cool peeps what they use. Well here's a list of interesting stuff:
# remote debuggers
- [qdb](https://github.com/quantopian/qdb) from quantopian
- [ptvsd](https://github.com/Microsoft/ptvsd/) the python debugger for VS from the microsoft
- [celery.contrib.rdb](https://docs.celeryproject.org/en/latest/userguide/debugging.html) from the celery peeps
- [python-remote-pdb](https://github.com/ionelmc/python-remote-pdb) from one of the `pytest` devs
- [epdb](https://github.com/sassoftware/epdb) and extended `pdb` - looks a little unmaintained tho
- [pdb++ discussion on better support for remote debugging](https://github.com/pdbpp/pdbpp/issues/351)

What I want:
- write `breakpoint()` in actor code and `tractor` does the right thing and yields std stream control in FIFO order to whatever actor got their message to the parent first
- first class support for [`pdbpp`](https://github.com/pdbpp/pdbpp), it's just too good not to use
- the ability to switch between multiple actors who have breakpoints in a sensible way much like [`gdb` does with threads](https://www.youtube.com/watch?v=aypNWcRlwYw)

As always, lurkers feel free to pipe in with ideas :surfing_man:

goodboy:
Been doing lots of digging as part of #129 and dumping some links:
- `ptk` has utils for [not clobbering stdout](https://github.com/prompt-toolkit/python-prompt-toolkit/blob/c2c6af8a0308f9e5d7c0e28cb8a02963fe0ce07a/prompt_toolkit/patch_stdout.py) which would be super handy for having multiple actors logging **while** your in the middle of debugging a crash.

More to come..

goodboy:
Woa found this project digging up profiler research, it's from the same depa/team(s) as [`scalene`](https://github.com/plasma-umass/scalene?tab=readme-ov-file#using-scalene),

- > ChatDBG - AI-assisted debugging. Uses AI to answer 'why' 

https://github.com/plasma-umass/ChatDBG?tab=readme-ov-file#examples

Issue #112, opened 2020-02-10
Settle the flaky tests..


goodboy:
On `mp` we'll get [hangs on the streaming tests](https://travis-ci.org/goodboy/tractor/jobs/648528123?utm_medium=notification&utm_source=github_status) and on `trip` we sometimes get [strange bails](https://travis-ci.org/goodboy/tractor/jobs/648528126?utm_medium=notification&utm_source=github_status) on the `test_nested_multierrors[trio_run_in_process]` test.

goodboy:
Some of this might have to do with #115..

goodboy:
This should be mostly getting resolved in #128.

goodboy:
Closed in #128.

PR #128, opened 2020-07-22
Drop trio-run-in-process,  use pure trio process spawner, test out of channel ctrl-c subactor cancellation

### LINKED ISSUE #117: Let's roll our own subproc spawner ("the way the experts would")

goodboy:
This continues from #127 since some history was reworked to get some upstream changes in before this to do with #121.

Much thanks to @guilledk for starting this effort!

The main things we have accomplished thus far:
- [x] drop `trio-run-in-process` in place of our own native `trio` based spawner thus solving #117
- [x] bypassing and thus resolving #115 (new tests verify this)

Things still on the *todo*:
- [x] (**solved via hand-crafted rebasing**) figure out if we can make this land before #121 (which still has a bunch of outstandings)
  - Techincally, there's no documentation of it so it might not hurt to just merge?
    - note to self: the only commits we probably want to hold off on are found with `git log --follow tractor/to_asyncio.py`
- [x] (**probably can defer for now, though we need a new issue**) determine if we want to drop `cloudpickle` and instead pass all `Actor.__init__()` data to `Actor._async_main()` allowing us to just use `msgpack` and the `Actor.load_modules()`/`Actor._get_rpc_func()` way of invoking the target entrypoint.
- [x] (**should have been resolved in #129**) discover if there's anything else related to remote debugging (#113) that can be addressed now as part of subproc spawning before we move on

Closes #117, #115,  #112


goodboy:
grrr,

`tractor/to_asyncio.py:47: error: Missing return statement`

This `mypy` stuff is killin me.

goodboy:
@guilledk @carlosplanchon this is now rebased onto the changes hand-pulled from the `infect_asyncio` branch and is ready for merge with master (I think).

Please take a look at this if you have the time :pleading_face:

goodboy:
Ok somewhow we've borked windows?
It looks like `pytest` is collecting but not running?
Gonna try pushing up a `.travis.yml` tweak.

The `multiprocessing` run on linux is shaky (as usual) so I wonder if we should go further with not pushing the nested subactors so much in those tests even if just for making the CI process a little smoother.

PR #129, opened 2020-07-23
Wen? Multiprocessing-native debugger now!

goodboy:
This is a first attempt at accomplishing #113 :open_mouth: 

Surprisingly it all works without using any threads, `greenback`, or other dark magic; it's just `tractor` bby :kiss:.

Hopefully we can get some feedback from lurkers :detective:

This first edition definitely works as well as other alternatives but there's a bunch of stuff we can do to make it better.

### Caveats:
- ~only first *level* actors can be debugged this way since the parent requires access to the subprocesse's stdin pipe~
  - any actor in a tree is now supported!
- ~2nd depth level actors are completely untested for now~
  - they are definitely now tested!
- no clue if this will work on windows (**and I even tried, see #158**)

### Immediate TODOs:
- [x] (**we'll need to defer this until #130 is adressed**) 1. disable the "cooked" mode when processing stdin (you'll see some commented code where I already started on this) such that each character is relayed immediately to the child stdin pipe.
  - This should get us out of the box completions and line editor control from the remote `pdbpp` :tada: 
  - I'm working off [this example code](https://www.programcreek.com/python/example/89186/tty.setcbreak); if there are terminal nerds about please feel free to yell about anything wrong with this approach :bowing_man
- [x] 2. lock access to the parent's stdin such that no more then one subactor can hijack it at a time; this of course is required in order to avoid all sorts of console mayhem
  - I've got locking working but not *switching* between two subactors requesting the parents stdio..
- [x] 3. somehow test all this? I think taking at look at how [`pytest` does it is probably a good start](https://github.com/pytest-dev/pytest/blob/master/testing/test_debugging.py).
- [x] 4. get feedback from some real users and figure out if `debug_mode` is cool or not.
  - thinking this will come with `.dev0` usage so marking complete for now.

### For the future:
- [x] make an issue summarizing all this: **see #130**
  - for multi-host systems we need a way to of course relay to stdin in the child over a `Channel`. There seems to be a variety of projects that do this mostly listed in #113. I want to use `pdb++` since I think it's the best UX available, but we may need to [hook into its `sys.stdin` in a sneaky way](https://github.com/pdbpp/pdbpp/blob/master/src/pdbpp.py#L1074).
    - `epdb` [seems to have code for this (on regular `pdb`)](https://github.com/sassoftware/epdb/blob/master/epdb/epdb_server.py#L251) but I haven't [tested it](https://github.com/pdbpp/pdbpp/issues/351#issuecomment-578440043).
  - we probably want python-trio/trio#174 to have a better api experience *while* hijacking the parent's stdin, also see:
    - the particular [strategy outline from that issue](https://github.com/python-trio/trio/issues/174#issuecomment-465098700)
    - python-trio/trio#1395
    - python-trio/trio#219
  - do we want to support overriding the [`breakpoint()` built-in](https://www.python.org/dev/peps/pep-0553/#sys-breakpoint) (I also have the draft code commented in here)
    - users will still have to `await breakpoint()` unless we implement it a different way


goodboy:
@guilledk If you get a chance try to play with this as well :)

goodboy:
Alright, so after much frustration trying to get tab completion to *just work* I went back to `trip` and found that it *did just work* and I finally got the :facepalm: that of course it's not going to work with our code since the child process isn't attached the parent's (nor any) tty :crying_cat_face: 

Of course [`rlcompleter`](https://github.com/python/cpython/blob/master/Lib/rlcompleter.py#L27) (and thus `pdbp` and `pdbp++`) rely on a tty at startup for any `readline` support which is what gives all the fancy tab completion and control character functionality. I'm pretty sure even somehow hacking a `pty` around `breakpoint()` calls isn't going to work either especially since there's a whole package ([`ptyprocess`](https://github.com/pexpect/ptyprocess/)) built to solve the problem of launching subprocs under ptys..

I think instead of doing a big writeup here I'm going to make a new issue and discuss my findings.
2. - 4. should still get complete before this lands.

guilledk:
Hey checked out this branch and ran the tests, noticed test_rpc.py was failing, it was a small fix but don't know if its too generic now, checkout
https://github.com/guilledk/tractor/commit/a5306bd786bb61f23f78586c0b0652b066d81787

goodboy:
@guilledk yeah, I'm not following that change?

The debugging stuff I don't think should have an effect here?

goodboy:
FYI we need #131 / #133 to get local process tab complete working (aka fancy debuggin stuff) which is what was in my last comment here and @guilledk is working on this currently :tada:

debugging/mp_debug.py
```diff
@@ -0,0 +1,31 @@
+import tractor
```
goodboy, line 1: This is just a manual test script for now.

setup.py
```diff
@@ -39,7 +39,7 @@
     ],
     install_requires=[
         'msgpack', 'trio>0.8', 'async_generator', 'colorlog', 'wrapt',
-        'trio_typing',
+        'trio_typing', 'pdbpp',
```
goodboy, line 42: This is actually required as a dependency since we needed something where we could hook into the "tear down hooks" (i.e. where a user releases the debugger and let's code run again using either of the `continue` or `quit` commands).

tractor/_debug.py
```diff
@@ -0,0 +1,208 @@
...
+# TODO: is there some way to determine this programatically?
+_pdb_exit_patterns = tuple(
+    str.encode(patt + "\n") for patt in ('c', 'cont', 'continue', 'q', 'quit')
+)
+
+
+_pdb_release_hook = None
+
+
+class PdbwTeardown(pdbpp.Pdb):
```
goodboy, line 39: These are the teardown hooks.

tractor/_debug.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+log = get_logger(__name__)
+
+
+__all__ = ['breakpoint', 'post_mortem']
+
+
+# TODO: is there some way to determine this programatically?
+_pdb_exit_patterns = tuple(
```
goodboy, line 31: This can be commented out now I think.

tractor/_debug.py
```diff
@@ -0,0 +1,208 @@
...
+    """Custom ``pdbpp`` goodness.
+    """
+    sticky_by_default = True
+
+    def teardown(self, _pdb):
+        _pdb_release_hook(_pdb)
+
+
+# override the pdbpp config with our coolio one
+pdbpp.Pdb.DefaultConfig = TractorConfig
```
goodboy, line 63: Ahh, this can probably be just assigned on the `class` def.

tractor/_debug.py
```diff
@@ -0,0 +1,208 @@
...
+
+#             # relay bytes to subproc over pipe
+#             # await proc.stdin.send_all(bmsg)
+
+#             if bmsg in _pdb_exit_patterns:
+#                 log.info("Closing stdin hijack")
+#                 break
+
+
+async def _hijack_stdin_relay_to_child(
```
goodboy, line 97: This is run in the "root most" actor.

It's really just a simple lock since we're just trying to prevent other subactors that call `pdb.set_trace()` from doing it until no others are.

tractor/_debug.py
```diff
@@ -0,0 +1,208 @@
...
+
+    finally:
+        if debug_lock.locked():
+            debug_lock.release()
+            log.debug(f"Actor {subactor_uid} released stdin hijack lock")
+
+
+# XXX: We only make this sync in case someone wants to
+# overload the ``breakpoint()`` built-in.
+def _breakpoint(debug_func) -> Awaitable[None]:
```
goodboy, line 130: This is called in the child.

We still need detection for recursive calls to `await tractor.breakpoint()` from the same actor otherwise we'll get a deadlock around the lock on the parent.

It should be as simple as checking if there's an actor local var set.

tractor/_debug.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+    actor = tractor.current_actor()
+    do_unlock = trio.Event()
+
+    async def wait_for_parent_stdin_hijack(
+        task_status=trio.TASK_STATUS_IGNORED
+    ):
+
+        # TODO: need a more robust check for the "root" actor
+        if actor._parent_chan:
```
goodboy, line 142: This is actually still problematic for actors more then 1 nursery *deep* since those actor's parent won't be attached to a tty. So really we need a way to contact the root most actor to lock **it**.

The easiest way to do this would likely be to pass the root data down to each subactor I guess and then ask the arbiter for the root actor in a tree?

tractor/_debug.py
```diff
@@ -0,0 +1,208 @@
...
+
+    log.debug(f"Actor {subactor_uid} is waiting on stdin hijack lock")
+    await debug_lock.acquire()
+    log.warning(f"Actor {subactor_uid} acquired stdin hijack lock")
+
+    # TODO: when we get to true remote debugging
+    # this will deliver stdin data
+    try:
+        # indicate to child that we've locked stdio
+        yield 'Locked'
```
goodboy, line 113: Notice this will be treated as a streaming function but we only use one `yield`.
This IMO is a perfect example of something well wrapped by a cross-actor context manager that can be wrapped by the portal api. We actually don't require any special new messages in the IAC protocol either since we can just use the existing `yield` and `stop` messages. This is pretty similar to how `@contextmanager` on a generator function is mapping `yield` and `StopIteration` to the `__enter__()`/`__exit__()` api.

tractor/_debug.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def unlock(_pdb):
+        do_unlock.set()
+
+    global _pdb_release_hook
+    _pdb_release_hook = unlock
+
+    async def _bp():
+        # this must be awaited by caller
+        await actor._root_nursery.start(
```
goodboy, line 174: We use the root nursery so as not to *block* the debugger by being limited to a new nursery scope.

Should this maybe be better commented?

tractor/_entry.py
```diff
@@ -57,6 +57,9 @@ def _trio_main(
 ) -> None:
     """Entry point for a `trio_run_in_process` subactor.
     """
+    # TODO: make a global func to set this or is it too hacky?
```
goodboy, line 60: I'm starting to think this might be a good thing to do in subactors to avoid users thinking they can debug with plain old `breakpoint()` (which will still engage the debugger but potentially cause all sorts of console clobbering because there is no locking around the parent tty).

tractor/_portal.py
```diff
@@ -324,23 +328,25 @@ async def open_portal(
...
-                # if the local task is cancelled we want to keep
-                # the msg loop running until our block ends
-                shield=True,
+        msg_loop_cs = None
+        if start_msg_loop:
+            msg_loop_cs: trio.CancelScope = await nursery.start(
+                partial(
+                    actor._process_messages,
+                    channel,
+                    # if the local task is cancelled we want to keep
```
goodboy, line 345: This is new to shut down logic: we always keep the parent channel message loop running until the very end.

tractor/_actor.py
```diff
@@ -744,11 +771,19 @@ async def cancel(self) -> None:
...
         await self.cancel_rpc_tasks()
         self.cancel_server()
         self._root_nursery.cancel_scope.cancel()
+        if self._parent_chan is None:  # TODO: more robust check
+            # we're the root actor or zombied
+            root = trio.lowlevel.current_root_task()
+            for n in root.child_nurseries:
+                n.cancel_scope.cancel()
+
+        self._parent_chan_cs.cancel()
```
goodboy, line 786: need and `if` check here.

tractor/_actor.py
```diff
@@ -694,6 +717,10 @@ async def _async_main(
             # or completed
             self.cancel_server()
 
+            # teardown msg loop with parent
+            if self._parent_chan_cs:
```
goodboy, line 721: This is new to shut down logic: we always keep the parent channel message loop running until the very end.

It is cancelled as the last thing before process termination.

goodboy:
Got almost fully desired test coverage now!

Feel free to try out the new examples in `examples/debugging` on this branch :surfer: 

Would appreciate some manual feedback before the last of it lands.

tractor/_debug.py
```diff
@@ -0,0 +1,321 @@
...
+        "If you have a special need for it please open an issue.\n"
+    )
+
+
+# don't allow those stdlib mofos to mess with sigint handler
+pdbpp.pdb.Pdb.sigint_handler = handler
+
+
+# @contextmanager
+# def _disable_sigint():
```
goodboy, line 149: Didn't' really end up needing this.

goodboy:
Alright this is pretty much in a usable state with tests and good enough for now to be trialed in real production code (hint: we've already been doing it).

There's one remaining issue that I'm still confused about that's now demonstrated in the [commit here](https://github.com/goodboy/tractor/commit/3917117965c4d68e1ad82728eabd30eb0d238e62) on the new `bug_in_debug` branch.

I don't know if it's a huge deal to solve that issue right now (though it is kind of a big deal in general?) and it may require further investigation into what happens in children which are executing CPU bound routines when they receive a cancel request. I'm thinking it has to be something else but best guess right now is that the first child is not effectively relaying a cancel to it's child which is in debug; tracing through the log messages seems to show this.

goodboy:
Worth mentioning is that I have not idea if this stuff works no windows since there doesn't seem to be any `pexpect` support?

Might have to dig into this so that likely means a deferral issue is coming.

tests/test_cancellation.py
```diff
@@ -413,3 +413,38 @@ async def main():
...
+    "Sync sleep."
+    time.sleep(period)
+
+
+async def spawn():
+    async with tractor.open_nursery() as tn:
+        portal = await tn.run_in_actor('sleeper', spin_for)
+
+
+@no_windows
```
goodboy, line 429: Probably need an issue for this as well.

tractor/_actor.py
```diff
@@ -171,13 +176,16 @@ class Actor:
     _root_n: Optional[trio.Nursery] = None
     _service_n: Optional[trio.Nursery] = None
     _server_n: Optional[trio.Nursery] = None
+    _lifetime_stack: ExitStack = ExitStack()
```
goodboy, line 179: Oh right, we might want tests for this :rofl:..

Maybe it's ok to just slip it in for now and worry later?

guilledk (approved):

tests/test_debugger.py
```diff
@@ -0,0 +1,362 @@
...
+from os import path
+
+import pytest
+import pexpect
+
+from conftest import repodir
+
+
+# TODO:
+# - recurrent entry to breakpoint() from single actor *after* and an
```
goodboy, line 16: Actually, I think most of these are covered now but wouldn't hurt to do an audit in the future.

I don't have much juice left to keep pushing here :crying_cat_face:

tests/test_debugger.py
```diff
@@ -0,0 +1,362 @@
...
+    before = str(child.before.decode())
+    assert "AssertionError" in before
+
+
+def test_multi_nested_subactors_error_through_nurseries(spawn):
+    """Verify deeply nested actors that error trigger debugger entries
+    at each level up the tree.
+    """
+
+    # TODO: inside this script there's still a bug where if the parent
```
goodboy, line 320: This is actually fixed now and is addressed by solving #156.

PR #133, opened 2020-07-25
Drop cloudpickle dependency

guilledk:
In the new trio backend, we were using cloudpickle to serialize the async function the child will have to pass to trio to run, but this function never changed, it was `Actor._async_main`, this function handles the connection and handshake with the parent actor.

The real problem was passing the subactor object instance to the newly spawned child.

`tractor._actor.Actor` init function params:
- name: `str`,
- rpc_module_paths: `List[str]`,
- uid: `str`,
- loglevel: `str`,
- arbiter_addr: `Tuple[str, int]`

This parameters are passed through the child shell arguments and parsed using `argparse`.

The problem is the following, when running `tests/_test.py` from the shell this is the output:

```
[guille@hardPC tractor]$ python tests/_test.py 
Jul 25 16:42:27 (no actor context, 161699, tractor._main)) [WARNING] tractor __init__.py:70 No actor could be found @ 127.0.0.1:1616
Jul 25 16:42:27 (no actor context, 161699, tractor._main)) [DEBUG] tractor _actor.py:209 arbiter Loaded RPC modules: {}
Jul 25 16:42:27 (arbiter, 161699, tractor._main)) [INFO] tractor _actor.py:897 Starting local <tractor._actor.Arbiter object at 0x7ff5b0c9ca90> @ 127.0.0.1:1616
Jul 25 16:42:27 (arbiter, 161699, tractor._actor.Actor._serve_forever)) [DEBUG] tractor _actor.py:695 Started tcp server(s) on [<trio.socket.socket fd=8, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('127.0.0.1', 1616)>]
Jul 25 16:42:27 (arbiter, 161699, tractor._actor.Actor._async_main)) [DEBUG] tractor _actor.py:602 Registering <tractor._actor.Arbiter object at 0x7ff5b0c9ca90> for role `arbiter` @ ('127.0.0.1', 1616)
Jul 25 16:42:27 (arbiter, 161699, tractor._actor.Actor._async_main)) [DEBUG] tractor _actor.py:611 Waiting on root nursery to complete
Jul 25 16:42:27 (arbiter, 161699, tractor._main)) [DEBUG] tractor _actor.py:209 __main__ Loaded RPC modules: {'__main__': '/home/guille/Documents/tractor/tests/_test.py'}
[...]

PR #138, opened 2020-07-29
Start up sequence trickery

goodboy:
This is duplicate of #133 but just trying to trigger CI again.


goodboy:
Sooo weird. It should be impossible for these tests to run?

goodboy:
@guilledk see if that skip logic works. If so, merge it into your branch and we'll land #133.

goodboy:
:tada:

Issue #141, opened 2020-08-03
Subactors don't de-register from arbiter on cancel?


goodboy:
I noticed while working on `piker` that we are leaking registry entries when a subactor is cancelled.
Need tests and a fix for this stat!

goodboy:
Figured it out: it's a regression from #128 which is actually my fault.
Turns out if you `__aexit__()` a `trio.Process` it'll kill the process on cancellation before actor teardown happens (which is not the "soft kill" we want).

Gonna add some tests for this now that I think I've got a fix.

goodboy:
Huh, on even further investigation it's only when subactors are cancelled and the root actor **is not** the arbiter that the problem arises; super interesting.

PR #143, opened 2020-08-03
Ensure actors de-register with arbiter when cancelled during infitinite streaming.

goodboy:
This is a (finally) reproduction of the issue I discovered manually in #141; it was likely a regression (caused by #128) but it's hard to know since we didn't have tests for this until now :smile:

It turns out the issue only appears if you are in the middle of consuming and infinite stream and are cancelled.
The cancellation causes the `trio.Process` backend to kill the process before it can cleanup resulting in stale entries in the arbiter registry.

This first set of commits will demonstrate the issue (that I want to see also in CI) and then I will push up the solution commit :surfer:

### DIFF
--- a/tractor/_spawn.py
+++ b/tractor/_spawn.py
@@ -157,7 +157,6 @@ async def cancel_on_completion(
 @asynccontextmanager
 async def spawn_subactor(
     subactor: 'Actor',
-    accept_addr: Tuple[str, int],
     parent_addr: Tuple[str, int],
 ):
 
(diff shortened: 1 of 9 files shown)


goodboy:
Alright there's the expected failures.
Pushing up the fix :smile_cat:

goodboy:
@guilledk just waiting for your "ok" I think :+1:

tractor/_spawn.py
```diff
@@ -167,8 +166,11 @@ async def spawn_subactor(
         # Hardcode this (instead of using ``_child.__name__`` to avoid a
         # double import warning: https://stackoverflow.com/a/45070583
         "tractor._child",
+        # This is merely an identifier for debugging purposes when
+        # viewing the process tree from the OS
```
guilledk, line 170: Not sure if this is accurate, I mean for sure seeing the uid in the proc tree is useful for debugging but, also its needed for the `tractor._child` initialization.

goodboy: Ahh, I think my point was that otherwise we'd just pass it over the channel from the parent.
Maybe I should clarify that *we only pass this on cmdline because..* ?

ryanhiebert: That sounds like it would be clearer.

goodboy: Fixed :smile_cat:

goodboy:
@ryanhiebert @guilledk appreciate the review when ready :)

PR #144, opened 2020-08-07
Fix for dereg failure on manual stream close

goodboy:
There was code from the last de-registration fix PR (#141, #143) that [I had commented](https://github.com/goodboy/tractor/blob/master/tractor/_actor.py#L681)
(to do with shielding arbiter dereg steps in `Actor._async_main()`) because
the block didn't seem to make a difference under infinite streaming
tests. 

Turns out it for sure is needed if you manually close the stream **before** cancelling the containing nursery.

Pushing up a test to demonstrate the failure then will push up a fix to correct.

**UPDATE**: this ended up being a much bigger change set see comment below.

- Long story short, getting things to work reliably led to some reworking of cancellation machinery in terms of internal nursery composition. 
- I tried bringing code from newer feature branches to solve this de-registration issue but realized there was too much raciness at teardown to even reproduce the issue easily. 
- This rework *should* ready the core for even more deterministic actor cancellation soon. Ideally we can fully shield a call to `Portal.cancel_actor()` without a timeout in the future. See the commit messages for some commentary on this.
- overall the runtime startup and teardown logic should be much easier to read and understand now.

### DIFF
--- a/tests/test_discovery.py
+++ b/tests/test_discovery.py
@@ -196,11 +196,13 @@ def test_subactors_unregister_on_cancel(
     """
     with pytest.raises(KeyboardInterrupt):
         tractor.run(
-            spawn_and_check_registry,
-            arb_addr,
-            use_signal,
-            False,  # remote arbiter
-            with_streaming,
+            partial(
+                spawn_and_check_registry,
+                arb_addr,
+                use_signal,
+                remote_arbiter=False,
+                with_streaming=with_streaming,
+            ),
             arbiter_addr=arb_addr
         )
 
@@ -220,11 +222,13 @@ def test_subactors_unregister_on_cancel_remote_daemon(
     """
     with pytest.raises(KeyboardInterrupt):
         tractor.run(
-            spawn_and_check_registry,
-            arb_addr,
-            use_signal,
-            True,  # remote arbiter
-            with_streaming,
+            partial(
+                spawn_and_check_registry,
+                arb_addr,
+                use_signal,
+                remote_arbiter=True,
+                with_streaming=with_streaming,
+            ),
             # XXX: required to use remote daemon!
             arbiter_addr=arb_addr
         )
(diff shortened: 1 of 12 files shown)


goodboy:
My apologies, this ended up turning into a much larger change set which hardens the teardown machinery quite a bit.

Basically I reorged the internal nursery scoping to keep the channel with the parent actor up as long as possible such that cancellation is more deterministic and we get way fewer race conditions. The CI should (hopefully) demonstrate this moving forward.

tests/test_discovery.py
```diff
@@ -223,8 +223,109 @@ def test_subactors_unregister_on_cancel_remote_daemon(
...
+                    n.start_soon(streamer, agen1)
+                    n.start_soon(cancel, use_signal, .5)
+                    try:
+                        await streamer(agen2)
+                    finally:
+                        # Kill the root nursery thus resulting in
+                        # normal arbiter channel ops to fail during
+                        # teardown. It doesn't seem like this is
+                        # reliably triggered by an external SIGINT.
+                        # tractor.current_actor()._root_nursery.cancel_scope.cancel()
```
goodboy, line 272: Probably needs to get deferred to a new test module:
`tests_hard_kill.py` or something.

goodboy:
Oof definitely some grammar issues in some commit messages :laughing:

tests/test_discovery.py
goodboy: Made #145.

tractor/_actor.py
```diff
@@ -540,6 +549,48 @@ async def _process_messages(
...
+                    setattr(self, attr, value)
+
+            return chan, accept_addr
+
+        except OSError:  # failed to connect
+            log.warning(
+                f"Failed to connect to parent @ {parent_addr},"
+                " closing server")
+            await self.cancel()
+            # self._parent_chan = None
```
goodboy, line 591: Missed this.

tests/test_discovery.py
```diff
@@ -223,16 +223,73 @@ def test_subactors_unregister_on_cancel_remote_daemon(
             spawn_and_check_registry,
             arb_addr,
             use_signal,
-            True,
+            True,  # remote arbiter
```
ryanhiebert, line 226: Opaque booleans make for great required keyword arguments.

goodboy: @ryanhiebert i think the only reason for this was avoiding a `funtools.partial`.

Maybe it's better to be clearer and not worry about the extra import tho?

ryanhiebert: I wouldn't worry about an extra import. Clarity beats that for sure. It's quite possible there's other worthwhile considerations, but an import seems like a concern to not have.

goodboy: lol nm it was already imported so you win :stuck_out_tongue_closed_eyes:

PR #151, opened 2020-08-31
Typelog sphinx theme wrong name on requirements.txt

guilledk:
Fixes bug introduced in #150.
Double gaf, wrong name on the `requirements-docs.txt` for the typelog sphinx theme.


guilledk:
This actually wasn't the cause of the error.

Issue #157, opened 2020-10-12
Revamp docs for .dev0 release


goodboy:
Now that #129 is about to land we need some sections on:
- [ ] using the debugger in a variety of process tree configurations
  - examples of both `.breakpoint()` and the crash handler system
  - example of how it's like `pytest`'s `--pdb` even with just plain old process local tasks
- [ ] adjustment of all examples to drop use of namespace prefixes when calling `Portal.run()` and `ActorNursery.run_in_actor()` once a fix for #69 lands
  - focus on *simplicity* of invoking remote functions with simple types as input
  - demonstrate how one should **not** rely on pickling objects through an example of state mutating?
- [ ] re-org all major sections into new pages, a rough outline might look like:
 - spawning basics with process tree snap shots for each example
 - cancellation and error propagation
 - debugging
   - examples of using logging to figure out how things went wrong if the debugger isn't enough
 - streaming and a real world example
 - process tree and messaging patterns
   - maybe should include some details about backends and gotchas?

goodboy:
The readme update in #163 also mentions maybe adding some diagrams for both messaging / protocols docs and supervision trees stuff:

>add some process tree diagrams using both something like [mermaid.io](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVERcbiAgICBBW0NocmlzdG1hc10gLS0-fEdldCBtb25leXwgQihHbyBzaG9wcGluZylcbiAgICBCIC0tPiBDe0xldCBtZSB0aGlua31cbiAgICBDIC0tPnxPbmV8IERbTGFwdG9wXVxuICAgIEMgLS0-fFR3b3wgRVtpUGhvbmVdXG4gICAgQyAtLT58VGhyZWV8IEZbZmE6ZmEtY2FyIENhcl1cbiAgICAgICAgICAgICIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In0sInVwZGF0ZUVkaXRvciI6ZmFsc2V9) and output from `pstree`?

goodboy:
There was a project i saw recently that had a sphinx theme I really loved; too bad can't find it again.

It had the automatic toc arrows pointing to the section on the left as you scrolled down the page.

ryanhiebert:
I'm not sure, but I've seen a couple projects using https://sphinx-book-theme.readthedocs.io/en/latest/, and I rather like it. Was thinking about using it for some internal documentation at work.

[...]

Issue #153, opened 2020-09-08
Add logo and github actions badge to readme


goodboy:
[Atomix does it](https://github.com/atomix/atomix#readme) using their own website but I think we can using web service [provided by github](https://stackoverflow.com/a/14494775)?

So should be something like:
 `"![alt text](https://github.com/[goodboy]/[tractor]/blob/[master]/logo.svg?raw=true)"`

The actions badge type and url can be figure out [here](https://actions-badge.atrox.dev/).

goodboy:
This might be the more adult way to do it with an *assets branch*?
https://stackoverflow.com/a/42677655

goodboy:
Actually, [`alakitty`](https://blog.christianduerr.com/), [does it even better using html](https://raw.githubusercontent.com/alacritty/alacritty/master/README.md).

goodboy:
Go this working with a plain old `.. image:` `ReST` directive in #163 and just putting our readme inside the docs dir (which I didn't realize you could do).

Also found this [blog post](https://olivermak.es/2016/01/svg-in-github-readme/) on sticking `.svg` files into GH readmes though didn't end up using it since it was markdown/html based which didn't seem to work when done in the `.rst` format.

So the logo part is closed by 61611ee and the actions badge part in bba47e4c7a2a.

ryanhiebert:
OK, that's cool. I didn't know the readme could be in the docs folder.

Issue #146, opened 2020-08-11
Process launching speedups: Roll our own "fork server" on *nix?


goodboy:
`mulitprocessing`'s [forkserver](https://github.com/python/cpython/blob/master/Lib/multiprocessing/forkserver.py) has proven servery unreliable for nested process launching (#84, #134) however it is *much much faster* then our current `trio` (aka async wrapped `subprocess`) spawning backend and, as such, we're keeping around for benchmarking purposes for when we eventually tackle the building our own internal subsystem for fast process launching on **linux**.

The issues with current unix forking apis is covered in quite good depth in @catern's [blog post](http://catern.com/posts/fork.html) and I think it's a multiple required read before diving into this.

I've had some (very) brief discussion and there's some ideas I'd like to investigate:
- benchmarking a spawner written with [`supervise_api`](https://github.com/catern/supervise)
  - the [test set](https://github.com/catern/supervise/blob/master/python/supervise_api/tests/test_supervise.py) is probably handy to read for digging into this
  - understanding the recommended use of [`sfork()`](https://github.com/catern/sfork) 
- looking at how we could build our own version of `mp`'s [fork request loop](https://github.com/python/cpython/blob/master/Lib/multiprocessing/forkserver.py#L254) using `sfork` which does minimal `tractor` templated process environment copying via a specific supervisor process

I'd love to get @catern's feedback on a sensible direction with the fast forking sub-system design as well :surfer:

goodboy:
Discussion surrounding this is now happening in python-trio/trio#1614 in terms of avoiding using a server all together which would actually be the most ideal.

goodboy:
Motivating blog post: https://pythonspeed.com/articles/python-multiprocessing/

goodboy:
Necro-bump i guess but it's never been a better time honestly to try and actually solve this given the new official sub-interpreters (sint) support landing in py 3.14 🕶 

https://docs.python.org/3.14/library/concurrent.interpreters.html#basic-usage

---

### a theory for solving the `os.fork()` runtime-state copy problem

- if you want process isolation, not just cpython-runtimes in sys threads, then i postulate it's likely possible to use fork from within a `trio.run()` by a combo of a process fork and a *sint spawn*:
  1. in parent process, start new sint for child prior to fork, fork the child
  1. in child, once aware it's a child, terminate parent sint (this is where most questions i have are), execute `trio.run()`, continue with actor runtime as normal,
[...]

PR #163, opened 2020-10-16
Readme rework draft

goodboy:
Check the [rendered version](https://github.com/goodboy/tractor/tree/readme_pump).

#### TODO:
- [x] a definitive example once #69 lands

- ~[ ] the equivalent output from `$TERM -e watch -n 0.1 "pstree -a $$" & python examples/parallelism/we_are_processes.py && kill $!`~ we stuck the command in the examples which I think is good enough for now.
  ~- [ ] maybe a `.gif` of the above bullet?~ deferred to #199
  
  Do we need the gifs?
  I'm fine to add them later if peeps insist.

#### Defer to "branding" PR?:
- [x] a logo re: #153
~- [ ] someone to pump our `.dev0` release somewhere (cuz I ain't doin it)~ defer to #2

#### Defer to docs rework PR (the readme should be simple)
~- [ ] add some process tree diagrams using both something like [mermaid.io](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVERcbiAgICBBW0NocmlzdG1hc10gLS0-fEdldCBtb25leXwgQihHbyBzaG9wcGluZylcbiAgICBCIC0tPiBDe0xldCBtZSB0aGlua31cbiAgICBDIC0tPnxPbmV8IERbTGFwdG9wXVxuICAgIEMgLS0-fFR3b3wgRVtpUGhvbmVdXG4gICAgQyAtLT58VGhyZWV8IEZbZmE6ZmEtY2FyIENhcl1cbiAgICAgICAgICAgICIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In0sInVwZGF0ZUVkaXRvciI6ZmFsc2V9) and output from `pstree`?~ deferred to #157


README.rst
```diff
@@ -1,6 +1,7 @@
 tractor
 =======
 A `structured concurrent`_, async-native "`actor model`_" built on trio_ and multiprocessing_.
```
ryanhiebert, line 3: This looks pretty good to me. If I were you, I'd want to decouple my pitch for tractor from `multiprocessing`, and leave that as an implementation detail.

ryanhiebert (approved):

README.rst
goodboy: @ryanhiebert it links to wikipedia not the stdlib; so it's in reference to actual muli-processing as an OS usage pattern.

Not sure if that was clear. Maybe *multi-processing* (with the dash) makes that more clear?

If you just mean that not focusing on it being implemented with multiple processes (since in theory we could somehow move to a threading system eventually?) then definitely worth considering.

ryanhiebert: Oh, I was thinking it was about the stdlib module. The dash would make more sense to me.

goodboy: Yeah, the plan is to move entirely off the stdlib module asap.

The only thing blocking that is #146 for which I'd like to keep the stdlib stuff to compare against.

guilledk (approved):

goodboy:
@ryanhiebert btw there's a matrix room now for the project if you feel like joining us.

goodboy:
I think I figured out we can make a pretty simple "worker pool" example to address the first todo once #69 comes.

Something along the lines of a round-robin request-response pool; real simple.

Just sticking this comment as reminder for meself.

tractor/to_asyncio.py
```diff
@@ -0,0 +1,181 @@
...
+            done_callback=trio_done_callback,
+        )
+
+        (await trio_done_fut).unwrap()
+
+    # might as well if it's installed.
+    try:
+        import uvloop
+        loop = uvloop.new_event_loop()
+        asyncio.set_event_loop(loop)
```
chrizzFTD, line 177: Minor, but the helper [`uvloop.install()`](https://github.com/MagicStack/uvloop#using-uvloop) exists. I'm unsure of any actual difference, though.

README.rst
```diff
@@ -13,16 +14,22 @@ A `structured concurrent`_, async-native "`actor model`_" built on trio_ and mul
...
-distributed multi-core Python; it aims to be the Python multi-processing
-framework *you always wanted*.
+``tractor`` is a `structured concurrent`_ "`actor model`_" built on trio_ and multiprocessing_.
+
+It is an attempt to pair trionic_ `structured concurrency`_ with
+distributed Python. You can think of it as a ``trio`` *-across-processes*
+or simply as a replacement for the stdlib's `multiprocessing` but built
+on async primitives for IPC.
+
+``tractor``'s nurseries lets you spawn ``trio`` *"actors"*: new Python
```
chrizzFTD, line 24: Super nitpicky, but I think `nurseries let you` (instead of `lets`) is the correct grammar?

README.rst
```diff
@@ -45,17 +52,28 @@ No PyPi release yet!
     pip install git+git://github.com/goodboy/tractor.git
 
 
+Alluring Features
+-----------------
+- Infinitely nesteable process trees
+- A built-in API for inter-process streaming
+- A (first ever?) "native" multi-core debugger for Python using `pdb++`_
```
chrizzFTD, line 59: Excited to try this soon!

chrizzFTD (approved):
Proposed README updates look good to me, thanks @goodboy !

I want to come back and try the examples and a bit more in the coming weeks on Windows 😃

README.rst
goodboy: :partying_face:

tractor/to_asyncio.py
goodboy: Thanks v gtk.

goodboy:
For the worker pool example I'm thinking we might as well re-do the [`concurrent.futures` prime number example](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor-example) so that peeps have a cross reference for what the *insides* of a `ProcessPoolExecutor` looks like implemented in `tractor`.

README.rst
goodboy: Pretty sure this is fixed now.

richardsheridan:
I guess I still have a number of issues with the current state of the explanations here and in the docs. You've got a great hook there with the tag line, but I submit that it may fall of the rails fairly quickly for green newcomers.

Let me propose a structure for your readme and see if we can even agree on that before going deeper. ✔️means you've covered this point, ❌means I think you need fresh text and explanation.

- Title and tagline✔️
- Introduction
    - What tractor is *functionally* (from the user's perspective) ❌ 
    - Core features, ordered from noob to expert level ❌
    - How it integrates with *my* existing code solution that needs async multiprocessing❌
        - This is probably the first place to mention trio or SC
- Examples ❌
    - Run one function
    - Worker Pool
    - Enter a debugger in a nested child process (see below)
- Install✔️
- Implementation and motivation ✔️
    - This is the **first time** you should mention the actor model, or why the project is even named `tractor`. 
- Help and Contributing ✔️

Why should nested child debugging be a README example? Because it's SO FLIPPIN COOL that it defies credibility, so demonstrating that you have a concise, friendly UX for it right away sells everything else that you are advertising in the project. And if your UX is not concise enough for this... fix it? Also some people use GIFs of REPL usage for this kind of thing, like [tqdm](https://github.com/tqdm/tqdm).

Feel free to go on as long as you want ranting in the implementation section about actors and SC. It's interesting! But it can't come up front or it may put newbies off. IMHO if the actors ever make an appearance in your basic examples or "standard" API, that is a minor failure to prevent implementation details from leaking.

Lastly, I think stuff like "don't be scared" or "it's just trio" is not that helpful to write. A well written intro will make these points clear implicitly!

Also don't forget to go put these kinds of changes into the stuff on readthedocs as well!

Unrelated, but your repository description should probably be the same tag line you have in the readme.

Hope this helps!!

goodboy:
@richardsheridan thanks for much for the input 😸 
I really appreciate the attention to detail and agree pretty much 100% with everything suggested.
Focus on the ergonomics, ease of use, delay technical jargon theory entirely.

> Feel free to go on as long as you want ranting in the implementation section about actors and SC. It's interesting! But it can't come up front or it may put newbies off. IMHO if the actors ever make an appearance in your basic examples or "standard" API, that is a minor failure to prevent implementation details from leaking.

Agreed. I wonder if instead of referring to each process + `trio` as an actor is something to keep hush?
Realistically the name can stand on it's own to some degree just as a word, heh?

> Lastly, I think stuff like "don't be scared" or "it's just trio" is not that helpful to write. A well written intro will make these points clear implicitly!

Noted. I think you're first 3 bullets are the most critical what will likely need some further scrutiny.
Hopefully we can get something simple to grok and efficient to read.

richardsheridan:
> I wonder if instead of referring to each process + `trio` as an actor is something to keep hush?
> Realistically the name can stand on it's own to some degree just as a word, heh?

I don't think you have to keep actors hush at all, just don't lead with that in the readme or docs. And over on gitter, it kind of came together for me when [you said a `trio.run` could be viewed as an actor behavior](https://gitter.im/python-trio/general?at=5fed18ed2084ee4b7880d4f2), but that wouldn't have been as useful to hear without a bunch of context.

goodboy:
Just added what I think will be a decent readme example. It's simple and we can demonstrate the basics with a proc monitor from terminal.

Give it a shot using
`$TERM -e watch -n 0.1  "pstree -a $$" & python examples/parallelism/we_are_processes.py || kill $!`

Interested to hear all your thoughts.

goodboy:
Found this which might be useful for avoiding example duplication in the readme:
https://github.com/bitranox/rst_include#usage

Have yet to try it though.

I'm thinking we include the `we_are_processes.py` example, show how to run and see the process tree using:
`$TERM -e watch -n 0.1 "pstree -a $$" & python examples/parallelism/we_are_processes.py && kill $!`, and then just reference the worker pool and point to the file. 

Oh, and make sure we nail all @richardsheridan's requirements, so I guess that means a bit of rewriting and also a debugger example. Hmm I guess a gif of that would be good.

examples/parallelism/concurrent_actors_primes.py
```diff
@@ -0,0 +1,117 @@
...
+                        value,
+                        portal
+                    )
+
+                # deliver results as they arrive
+                for _ in range(len(sequence)):
+                    yield await recv_chan.receive()
+
+        # deliver the parallel "worker mapper" to user code
+        yield _map
```
richardsheridan, line 95: This yield should be in a try block and the subsequent teardown should be in a "finally" block otherwise if an error is raised in the worker context block the workers may not shut down (depending on the semantics of the actor nursery!)

examples/parallelism/concurrent_actors_primes.py
```diff
@@ -0,0 +1,117 @@
...
+            # this starts a new sub-actor (process + trio runtime) and
+            # stores it's "portal" for later use to "submit jobs" (ugh).
+            portals.append(
+                await tn.start_actor(
+                    f'worker_{i}',
+                    enable_modules=[__name__],
+                )
+            )
+
+        async def _map(
```
richardsheridan, line 71: It seems odd to me to use an async generator here when you could use the nursery inside it to span the async context manager's yield statement and just yield `recv_chan` and a new `submit_job_chan` or something like that.

To make this work, `snd_chan` should be closed when the jobs are exhausted, so each `send_result` should receive a clone of `snd_chan` and do the `async with snd_chan: await snd_chan.send(...)` trick.

examples/parallelism/concurrent_actors_primes.py
```diff
@@ -0,0 +1,117 @@
...
+        await tn.cancel()
+
+
+async def main():
+
+    async with worker_pool() as actor_map:
+
+        start = time.time()
+        # for number, prime in zip(PRIMES, executor.map(is_prime, PRIMES)):
+        async for number, prime in actor_map(is_prime, PRIMES):
```
richardsheridan, line 107: OK i complained about the weirdness of the async generator above but this is api is not shabby.

examples/parallelism/concurrent_actors_primes.py
richardsheridan: At least suggest that the user wrap the async generator with `aclosing` so it's less footgunny!

examples/parallelism/concurrent_actors_primes.py
goodboy: Yah that's what I thought as well.

examples/parallelism/concurrent_actors_primes.py
goodboy: Ah good call.

goodboy: > To make this work, snd_chan should be closed when the jobs are exhausted, so each send_result should receive a clone of snd_chan and do the async with snd_chan: await snd_chan.send(...) trick.

This would also get us the equivalent of [`Executor.submit()`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.submit) I think?

examples/parallelism/concurrent_actors_primes.py
goodboy: @richardsheridan heh, so yeah actually the `finally` isn't technically needed since the nursery will 💯 tear everything down correctly - no zombies allowed 😸 

The `await tn.cancel()` is actually just there because without it the *daemon* actors will run forever in the background and the nursery will block. You basically use the daemon style actors when you intend for the "service" to run forever.

goodboy:
See  altendky/qtrio#221 for some similar work to what I was trying to accomplish with those fast few commits.

Also worth following to get this all working with GH readmes:
sphinx-contrib/restbuilder#25

goodboy:
Yeah, so I'm thinkin the worker pool example is just too lengthy to have in the readme.
See it here:
https://github.com/goodboy/tractor/tree/readme_pump#the-example-youre-probably-after

I'm thinking it's probably better to just link to it and the stdlib example and show a command to run it with the terminal process monitor?

I think a sub-proc debugger example and then the changes @richardsheridan suggested and we should be good?

goodboy:
Alright, new draft is [live here](https://github.com/goodboy/tractor/tree/readme_pump).

I still have to go through and do the rewrite / rejig based on @richardsheridan's [suggestions](https://github.com/goodboy/tractor/pull/163#issuecomment-753001792).

We're getting close 🏄🏼

docs/README.rst
```diff
@@ -45,17 +59,156 @@ No PyPi release yet!
...
+
+    $TERM -e watch -n 0.1  "pstree -a $$" \
+        & python examples/parallelism/concurrent_actors_primes.py \
+        && kill $!
+
+This uses no extra threads, fancy semaphores or futures; all we need
+is ``tractor``'s IPC!
+
+
+.. _concurrent_actors_primes: https://github.com/goodboy/tractor/blob/readme_pump/examples/parallelism/concurrent_actors_primes.py
```
goodboy, line 142: Ugh, this link is wrong.

docs/README.rst
```diff
@@ -1,28 +1,42 @@
...
-framework *you always wanted*.
+``tractor`` is a `structured concurrent`_ "`actor model`_" built on trio_ and multi-processing_.
+
+It is an attempt to pair trionic_ `structured concurrency`_ with
+distributed Python. You can think of it as a ``trio``
+*-across-processes* or simply as an opinionated replacement for the
+stdlib's ``multiprocessing`` but built on async programming primitives
+from the ground up.
+
+Don't be scared off by this description. ``tractor`` **is just ``trio``**
```
richardsheridan, line 25: This bold mixed with default role stuff doesn't render properly on github

goodboy: Yah, need to definitely fix this plus stick in your other suggestions.

Thanks again for the thoroughness.

goodboy: I might actually delay branding updates until a new PR since this is getting a bit out of hand 😂

goodboy:
Latest. hopefully final rework [is up](https://github.com/goodboy/tractor/tree/readme_pump)!

docs/README.rst
```diff
@@ -1,41 +1,170 @@
...
+- (Soon to land) typed messaging protocols (ex. via ``msgspec``)
+
+
+Zombie safe: self-destruct a process tree
+-----------------------------------------
+``tractor`` (attempts to) protect from zombies, no matter what.
+
+.. code:: python
+
+    """
```
McSinyx, line 40: Per the reST specs,

> Several constructs begin with a marker, and the body of the construct
> must be indented relative to the marker.

where in this case, the `code-block` (I can't seem to find `code` even in Sphinx-flavored reST) body should be indented by 3 space.  I believe that the editor picked this up and caused some indentation issues below.

docs/README.rst
```diff
@@ -1,41 +1,170 @@
...
+            f"Yo, i'm '{tractor.current_actor().name}' "
+            f"running in pid {os.getpid()}"
+        )
+
+       await trio.sleep_forever()
+
+
+    async def main():
+
+       async with tractor.open_nursery() as n:
```
McSinyx, line 66: Like here.

docs/README.rst
```diff
@@ -1,41 +1,170 @@
...
+
+           print('This process tree will self-destruct in 1 sec...')
+           await trio.sleep(1)
+
+           # you could have done this yourself
+           raise Exception('Self Destructed')
+
+
+    if __name__ == '__main__':
+       try:
```
McSinyx, line 79: Here as well.

McSinyx (commented):
Coming from the trio room on Matrix: I was wrong about the empty line after the headings though.

docs/README.rst
goodboy: Oh so you mean *after` the `..` there should be 3 spaces before the `code::`?

docs/README.rst
goodboy: Nice catch. Fixing all this.

goodboy:
@guilledk yeah not sure what's going on with that worker pool link.
Code looks right now?

docs/README.rst
```diff
@@ -1,41 +1,170 @@
...
+
+    $TERM -e watch -n 0.1  "pstree -a $$" \
+        & python examples/parallelism/concurrent_actors_primes.py \
+        && kill $!
+
+This uses no extra threads, fancy semaphores or futures; all we need
+is ``tractor``'s IPC!
+
+
+.. _concurrent_actors_primes: https://github.com/goodboy/tractor/blob/master/examples/parallelism/concurrent_actors_primes.py
```
goodboy, line 166: @guilledk not sure why it's not working?

goodboy: This might be it:
https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#use-custom-link-text

> Do not include a space between the last word of the link text and the opening angle bracket for the anchor text.

goodboy: Lol nope nm.

goodboy: Yeah no idea. The code looks right -> seems to be some GH thing.

docs/README.rst
McSinyx: Yep, e.g.

```rst
.. code-block:: python

   """
   ...
   async def target():
       try:
   ...
```

BTW could you please hide the resolved reviews, my international connection is terrible due to cable maintenance or something and GitHub's way of fetching content really does not help :smile:

goodboy:
Lol, well turns out CI did in fact find a bug due to #185.

It appears that `debug_mode=True` isn't triggering crash handling in the root process **only when** not explicitly using the `async with open_root_actor()` entry-point; with the implicit runtime, errors caught in the root actor aren't being bubbled into the wrapping `AsyncExitStack` when using `async with tractor.open_nursery()`.

I'm going to pull out the commit that introduced this change to the debugging example, move that history to a new branch + PR along with changes to port all the debugging examples to the new API.

PR #164, opened 2020-11-03
Drop duplicate project-package name in msg header

goodboy:
For projects that use the internal pretty logging system (such as `piker`) it's annoying to see the project-package name duplicated in log msg headers when using `tractor.log.get_logger(__name__)` from inside the client module:

prefer `piker.mod.submod submod.py:439` vs `piker.piker.mod.submod submod.py:439`

This just strips that duplicate token if detected while still keep the child logging hierarchy in tact.

### DIFF
--- a/tractor/log.py
+++ b/tractor/log.py
@@ -55,7 +55,15 @@ def get_logger(
     '''Return the package log or a sub-log for `name` if provided.
     '''
     log = rlog = logging.getLogger(_root_name)
+
     if name and name != _proj_name:
+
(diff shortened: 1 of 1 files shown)


ryanhiebert (commented):
I'm not really sure what to think on this, because I'm not sure what the `proj_name` is. As long as the logging path is understandable and unambiguous, this seems fine.

goodboy:
Probably should add a test, but been running it daily with no issues.

PR #165, opened 2020-11-16
Ignore SIGINT when in a debugger REPL

goodboy:
Some fixes for handling `SIGINT` / control-c when running in ~guest~ debug mode and dealing with `pdb` internals that [swap out the `sigint` handler from under our feet](https://github.com/python/cpython/blob/master/Lib/pdb.py#L344).

~This resolves an issue when running the `Qt` loop in guest mode where control-c *after* a `tractor.breakpoint()` was resulting in a hang. There were 2 problems:~
~- our [handler](https://github.com/goodboy/tractor/blob/master/tractor/_debug.py#L135) was borked (bad type signature)~
~- if we receive `SIGINT` in the root process we should always tear down the actor tree~

There's been a long-standing bug to do with the root actor getting stuck hanging for the tty lock to release after a child locks it then becomes uncontactable over IPC.

Part of that was due to:
- root and/or children accepting SIGINT (normally by accident since why would you want ctrl-c to work other then exiting the `pdb` repl while in the middle of it) **while in the debugger** which can trigger cancellation that will clobber the tty state
- the root process always handling SIGINT when a child is in debug -> likely you want to wait until the child is done (especially if it's ignoring SIGINT while debugging) before handling it in the root and cancelling any actor nurseries..

-------
still ToFix:
**deferred to #320**

---
still ToDo: **deferred to #320**

---
Next up:
- [x] ctl-c inside debugger REPL tests (for those that work)
  - [x] single depth tree actor and root-only level tests:
  ```
  tests/test_debugger.py::test_root_actor_bp_forever[ctl-c=True-trio]
  tests/test_debugger.py::test_subactor_error[ctl-c=True-trio]
  tests/test_debugger.py::test_subactor_breakpoint[ctl-c=True-trio]
  tests/test_debugger.py::test_multi_subactors[ctl-c=True-trio]
  tests/test_debugger.py::test_multi_daemon_subactors[ctl-c=True-trio]
  tests/test_debugger.py::test_multi_subactors_root_errors[ctl-c=True-trio]
  tests/test_debugger.py::test_root_nursery_cancels_before_child_releases_tty_lock[ctl-c=True-trio]
  tests/test_debugger.py::test_root_cancels_child_context_during_startup[ctl-c=True-trio]
  tests/test_debugger.py::test_different_debug_mode_per_actor[ctl-c=True-trio]
  ```

- [x] follow up issue (set) for what we need to solve in terms of the much harder cases  (#320)
  - [x] should we write expected to fail tests for these? **skipped em instead bc `xfail` wasn't workin**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,7 @@ jobs:
         run: pip install -U . -r requirements-test.txt -r requirements-docs.txt --upgrade-strategy eager
 
       - name: Run tests
-        run: pytest tests/ --spawn-backend=${{ matrix.spawn_backend }} -rs -v
+        run: pytest tests/ --spawn-backend=${{ matrix.spawn_backend }} -rsx
 
   # We skip 3.10 on windows for now due to
   # https://github.com/pytest-dev/pytest/issues/8733
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,7 @@ jobs:
         run: pip install -U . -r requirements-test.txt -r requirements-docs.txt --upgrade-strategy eager
 
       - name: Run tests
-        run: pytest tests/ --spawn-backend=${{ matrix.spawn_backend }} -rs -v --full-trace
+        run: pytest tests/ --spawn-backend=${{ matrix.spawn_backend }} -rs -v
 
   # We skip 3.10 on windows for now due to
   # https://github.com/pytest-dev/pytest/issues/8733
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,7 @@ jobs:
         run: pip install -U . -r requirements-test.txt -r requirements-docs.txt --upgrade-strategy eager
 
       - name: Run tests
-        run: pytest tests/ --spawn-backend=${{ matrix.spawn_backend }} -rs
+        run: pytest tests/ --spawn-backend=${{ matrix.spawn_backend }} -rs -v --full-trace
 
   # We skip 3.10 on windows for now due to
   # https://github.com/pytest-dev/pytest/issues/8733
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,9 @@ jobs:
       - name: Install dependencies
         run: pip install -U . -r requirements-test.txt -r requirements-docs.txt --upgrade-strategy eager
 
+      - name: List dependencies
+        run: pip freeze
+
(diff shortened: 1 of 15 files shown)


goodboy:
Pretty sure this get's defunct-ed by  #170 but leaving up temporarily in case something here is needed for the new `asyncio` branch (which still has this history iirc).

goodboy:
Still needs the [handler to be set](https://docs.python.org/3/library/signal.html#example)...

And also, oddly this code seems to be on master?

goodboy:
Ok found an odd one while testing on 3.10. Basically if you don't have `pdbpp` master you'll end up with this bt on a traceback inside a `@tractor.context` block where you call `.breakpoint()` and then hit a remote error:

```python
    File ".../tractor/_debug.py", line 729, in _post_mortem
      for _ in range(100):
    File ".../tractor/310/lib/python3.10/site-packages/pdb.py", line 1227, in xpm
      post_mortem(info[2], Pdb)
    File "...tractor/310/lib/python3.10/site-packages/pdb.py", line 1175, in post_mortem
      p.interaction(None, t)
    File "...tractor/310/lib/python3.10/site-packages/pdb.py", line 216, in interaction
      ret = self.setup(frame, traceback)
    File "...tractor/310/lib/python3.10/site-packages/pdb.py", line 259, in setup
      ret = super(Pdb, self).setup(frame, tb)
    File "/usr/lib/python3.10/pdb.py", line 217, in setup
      self.curframe = self.stack[self.curindex][0]
    IndexError: list index out of range
```
Pinning to `pdbpp`'s b757794857 seems to solve this 😎.
We probably should report issue that we're waiting on a release with this.

goodboy:
heyooo CI passss...

I guess we're getting close?

Next up:
- [ ] ctl-c inside debugger REPL tests (for those that work)
- [ ] follow up issue (set) for what we need to solve in terms of the much harder cases

IMO this PR has been delayed way more then long enough and we might as well push out another alpha to get this in as well as #293

goodboy:
Bleh, looks like `pdbpp` master and py3.9 aren't playing well..

goodboy:
Somehow this run passed: https://github.com/goodboy/tractor/runs/7063206690?check_suite_focus=true

But after that it doesn't?
I'm pretty lost on this with these weird timeouts now..

goodboy:
Lol, ok then.

goodboy:
Well then, that took just about long enough 😂

goodboy:
Squashed all the squashables, just needs nooz and I think we're good to land.

guilledk (approved):

PR #169, opened 2020-12-10
Py3.9

goodboy:
Added new version to CI.

Likely we'll have some warts to work through with the `multiprocessing.fork_server` stuff as usual...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python: ['3.7', '3.8']
+        python: ['3.7', '3.8', '3.9']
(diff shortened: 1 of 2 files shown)


goodboy:
Hmm I guess we should run `mypy` under 3.9 as well?

goodboy:
Heh, [of course it's the forkserver](https://github.com/goodboy/tractor/pull/169/checks?check_run_id=1532735445#step:5:16)...

goodboy:
This that logic guard will fix CI but should probably still go through the [recent history](https://github.com/python/cpython/commits/65a796e5272f61b42792d3a8c69686558c1872c5/Lib/multiprocessing/forkserver.py) and get whatever improvements we need.

If we're lucky maybe they've fixed some of the stuff we're cranky about in #146?

goodboy:
Well went through that history and looks like we have it all so I think this is good to go!

PR #170, opened 2020-12-11
End the `pdb` SIGINT handling madness

goodboy:
Turns out this is a lower level issue in terms of the stdlib's default
`pdb.Pdb` settings and how they conflict with `trio`s cancellation and
KBI handling. The details are hashed out more thoroughly in
python-trio/trio#1155. Maybe we can get a fix in trio so things are
solved under our feet :)

### DIFF
--- a/tractor/__init__.py
+++ b/tractor/__init__.py
@@ -117,9 +117,13 @@ async def _main(
     # Note that if the current actor is the arbiter it is desirable
     # for it to stay up indefinitely until a re-election process has
     # taken place - which is not implemented yet FYI).
-    return await _start_actor(
-        actor, main, host, port, arbiter_addr=arbiter_addr
-    )
+
+    try:
+        return await _start_actor(
+            actor, main, host, port, arbiter_addr=arbiter_addr
+        )
+    finally:
+        logger.info("Root actor terminated")
 
 
 def run(
--- a/tractor/_actor.py
+++ b/tractor/_actor.py
@@ -821,7 +821,7 @@ class Actor:
         self._server_down = trio.Event()
         try:
             async with trio.open_nursery() as server_n:
-                listeners: List[trio.abc.Listener] = await server_n.start(
+                l: List[trio.abc.Listener] = await server_n.start(
                     partial(
                         trio.serve_tcp,
                         self._stream_handler,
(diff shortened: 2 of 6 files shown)


tractor/__init__.py
```diff
@@ -117,9 +117,13 @@ async def _main(
...
-    return await _start_actor(
-        actor, main, host, port, arbiter_addr=arbiter_addr
-    )
+
+    try:
+        return await _start_actor(
+            actor, main, host, port, arbiter_addr=arbiter_addr
+        )
+    finally:
+        logger.info("Root actor terminated")
```
goodboy, line 126: These are probably superfluous (and maybe should be `.debug()`) but figured can't hurt atm. Likely we'll throttle back logging noise before first `.dev0`.

tractor/_actor.py
```diff
@@ -832,9 +832,10 @@ async def _serve_forever(
                         host=accept_host,
                     )
                 )
-                log.debug("Started tcp server(s) on"  # type: ignore
-                          f" {[l.socket for l in listeners]}")
-                self._listeners.extend(listeners)
+                log.debug(
+                    "Started tcp server(s) on"
+                    f" {[getattr(l, 'socket', 'unknown socket') for l in l]}")
```
goodboy, line 837: just `mypy` stuff since technically `.socket` isn't discovered through the type system easily.

tractor/_debug.py
```diff
@@ -132,19 +130,6 @@ async def _acquire_debug_lock(uid: Tuple[str, str]) -> AsyncIterator[None]:
         log.error(f"TTY lock released by {task_name}:{uid}")
 
 
-def handler(signum, frame):
```
goodboy, line 135: Turns out we can just keep `trio` standard KBI handling as long as we prevent `pdb.Pdb` from overriding it!

tractor/_debug.py
```diff
@@ -269,14 +254,29 @@ async def _bp():
...
     # user code **must** await this!
     return _bp()
 
 
+def _mk_pdb():
+    # XXX: setting these flags on the pdb instance are absolutely
+    # critical to having ctrl-c work in the ``trio`` standard way!
+    # The stdlib's pdb supports entering the current sync frame
+    # on a SIGINT, with ``trio`` we pretty much never want this
+    # and we did we can handle it in the ``tractor`` task runtime.
```
goodboy, line 266: should be *and **if** we did*..

guilledk (approved):

Issue #66, opened 2019-03-10
Worker pool super-system?


tgoodlet:
@parity3 made a gitter request for a system to delegate work to an actor/task cluster or *pool*.

>for my use case, I've got multiple "clients" that wish to use the same
set of workers, coming in at sporadic times. So they'd need to connect
to a running arbiter, and the arbiter needs to make sure that some
workers are free before starting on the client-specified set of task
chunks (as a method of resource manageement). It'd be nice to see an
example for that use case, or more built in api support for such case.

And with a little follow up requirement:
>also, it's not clear to me how best to implement a pattern where actors
are all notified that there are no more unclaimed tasks of a set (for
the actors that were involved with the task set). It would be nice for
such actors to be notified so they can "wrap up" any transient data
associated with the tasks they performed, aggregate / summarize it, and
send it back to the client

That last part will probably be to do with a lack of docs on how actor cancellation can be done using one of `ActorNursery.cancel()`, `Portal.cancel_actor()` or a plain old `trio.CancelScope`.

This *worker pool* is something I've though about a bit and it reminded me of a couple projects:
- the [routers and pools in `akka`](https://doc.akka.io/docs/akka/current/routing.html)
- the [*monitors* in `pulsar`](http://quantmind.github.io/pulsar/design.html#design-monitor)
- [worker pools in `erlang` OTP](https://learnyousomeerlang.com/building-applications-with-otp) for which there's a nice wrapped implementation names [`poolboy`](https://github.com/devinus/poolboy)
- an interesting post on the [need for a worker pool in `rabbitmq`](https://tech.labs.oliverwyman.com/blog/2010/03/29/on-the-limits-of-concurrency-worker-pools-in-erlang/) (introduced some time ago: 2010)

goodboy:
A much delayed follow up: we'll probably start with an example in the `README` / docs (#163) and add a more extensive tutorial surrounding this later on.

In terms of an API being incorporated into `tractor` itself I'd like to see a slew of implementations before getting to set on anything.

goodboy:
some new-gen competitors to [`concurrent.futures`](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor-example),
- `anyio`: https://github.com/Vizonex/anyio-executor
- `asyncio`: https://github.com/aio-libs/aiolibs-executor

PR #171, opened 2020-12-17
Add a way to shield a stream's underlying channel

goodboy:
Add a ``tractor._portal.StreamReceiveChannel.shield_channel()`` context manager which allows for avoiding the closing of an IPC stream's underlying channel for the purposes of task re-spawning. 

Sometimes you might want to cancel a task consuming a stream but not tear down the IPC
between actors (the default). A common use can might be where the task's "setup" work might need to be redone but you want to keep the established portal / channel in tact despite the task restart.

Includes a test.

I'm still not sure on the name here..

Any opinions @salotz , @guilledk?

Note it's used inside a `with stream.<meth name>`

Short list:
- `.shield_channel()`
- `.shielded_channel()`
- `.shield()`
- just make a `bool` property `.shield` you assign manually?
- `.shield_from_close()`
- `.channel_kept_open()`
- `.channel_shielded()`

Thoughts appreciated.

### DIFF
--- a/tests/test_streaming.py
+++ b/tests/test_streaming.py
@@ -7,6 +7,7 @@ import platform
 
 import trio
 import tractor
+from tractor.testing import tractor_test
 import pytest
 
 
@@ -53,6 +54,7 @@ async def stream_from_single_subactor(stream_func_name):
     """Verify we can spawn a daemon actor and retrieve streamed data.
     """
     async with tractor.find_actor('streamerd') as portals:
+
         if not portals:
             # only one per host address, spawns an actor if None
             async with tractor.open_nursery() as nursery:
@@ -73,8 +75,10 @@ async def stream_from_single_subactor(stream_func_name):
                 # it'd sure be nice to have an asyncitertools here...
                 iseq = iter(seq)
                 ival = next(iseq)
+
                 async for val in stream:
                     assert val == ival
+
                     try:
                         ival = next(iseq)
                     except StopIteration:
@@ -83,6 +87,7 @@ async def stream_from_single_subactor(stream_func_name):
                         await stream.aclose()
 
                 await trio.sleep(0.3)
+
                 try:
                     await stream.__anext__()
                 except StopAsyncIteration:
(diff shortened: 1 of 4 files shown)


tractor/_portal.py
```diff
@@ -83,19 +85,37 @@ async def receive(self):
                 "Received internal error at portal?")
             raise unpack_error(msg, self._portal.channel)
 
+    @contextmanager
+    def shield_channel(
+        self
+    ) -> typing.AsyncGenerator['StreamReceiveChannel', None]:
```
goodboy, line 91: Lol this type sig is [obv wrong](https://github.com/goodboy/tractor/pull/171/checks?check_run_id=1568199138); originally made it async, for no good reason..

tests/test_streaming.py
```diff
@@ -246,3 +254,68 @@ def test_not_fast_enough_quad(
...
+
+    This also serves to verify that all values from the stream can be
+    received despite the respawns.
+
+    """
+    stream = None
+
+    async with tractor.open_nursery() as n:
+
+        stream = await(await n.run_in_actor(
```
goodboy, line 277: Btw, what do y'all think about a `ActorNursery.stream_from_actor()` sugar for this pattern?

guilledk:
I prefer one of these:

- `.shield_channel()`
- `.shielded_channel()`
- `.shield()`

tests/test_streaming.py
goodboy: In theory you can actually just call `Portal.result()` before returning from `.run_in_actor()` but it means each call is then blocking and you'd have to spawn multiple (one-off) actors using tasks. Not sure if this is a problem and maybe will get hashed out with the discussion in #66.

goodboy:
Sticking with `.shield()` for now; can always change it later.

goodboy:
Ok think this is ready finally (after wrangling some tests code along the way).

I'll probably merge in the morning if noone pipes up.

PR #174, opened 2020-12-21
Allow passing function refs to `Portal.run()`

### LINKED ISSUE #69: Allow passing function refs to Portal.run() ?

goodboy:
This resolves and completes #69 allowing all RPC invocation APIs to pass
function references directly instead of explicit `str` names for the
target namespace and function (this is still done implicitly
underneath).  This brings us closer to `trio`'s task running API as well
as acknowledges that any inter-host RPC system (and API) will likely
need to be implemented on top of local RPC primitives anyway. Even if
this ends up **not** being true we can always go to "function stubs" as
part of our IAC protocol or, add a new method to do explicit namespace
calls: `.run_from_module()` or whatever everyone votes on.

Resolves #69

Further, this commit drops `Actor.statespace` from the entire system
since a user can easily get this same functionality using module
level variables. Fix docs to match all these changes (luckily mostly
already done due to example scripts referencing).

### DIFF
--- a/examples/actor_spawning_and_causality.py
+++ b/examples/actor_spawning_and_causality.py
@@ -2,7 +2,8 @@ import tractor
 
 
 def cellar_door():
-   return "Dang that's beautiful"
+    assert not tractor.is_root_process()
+    return "Dang that's beautiful"
 
 
 async def main():
@@ -10,7 +11,10 @@ async def main():
     """
     async with tractor.open_nursery() as n:
 
-        portal = await n.run_in_actor('some_linguist', cellar_door)
+        portal = await n.run_in_actor(
+            cellar_door,
+            name='some_linguist',
+        )
 
     # The ``async with`` will unblock here since the 'some_linguist'
     # actor has completed its main task ``cellar_door``.
--- a/tractor/_trionics.py
+++ b/tractor/_trionics.py
@@ -52,7 +52,7 @@ class ActorNursery:
         name: str,
         *,
         bind_addr: Tuple[str, int] = _default_bind_addr,
-        statespace: Optional[Dict[str, Any]] = None,
+        # statespace: Optional[Dict[str, Any]] = None,
         rpc_module_paths: List[str] = None,
         loglevel: str = None,  # set log level per subactor
         nursery: trio.Nursery = None,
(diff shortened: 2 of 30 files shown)


goodboy:
Heh, just realizing that `.run_in_actor()` will always cause the deprecation warning now smh.

I guess maybe moving what *was* `Portal.run()` to -> `._run()` will work for now.

guilledk (approved):

goodboy:
@guilledk I added back a `Portal.run_from_ns()` to keep the old api mostly for internal cross-actor method calls.

I see no harm in keeping this interface (since that's how it works underneath anyway) but we just won't publicize it in the docs.

goodboy:
Yo @ryanhiebert you mind a quick peek at this?

You're technically mostly to blame for this beautiful improvement 😝

examples/actor_spawning_and_causality.py
```diff
@@ -2,15 +2,19 @@
...
 async def main():
     """The main ``tractor`` routine.
     """
     async with tractor.open_nursery() as n:
 
-        portal = await n.run_in_actor('some_linguist', cellar_door)
+        portal = await n.run_in_actor(
+            cellar_door,
+            name='some_linguist',
+        )
```
ryanhiebert, line 17: I appreciate this different ordering of the arguments. It strikes my sensibilities more favorably.

tractor/_trionics.py
```diff
@@ -111,11 +111,16 @@ async def run_in_actor(
...
+        if name is None:
+            # use the explicit function name if not provided
+            name = fn.__name__
+
         portal = await self.start_actor(
             name,
             rpc_module_paths=[mod_path] + (rpc_module_paths or []),
             bind_addr=bind_addr,
-            statespace=statespace,
+            # statespace=statespace,
```
ryanhiebert, line 123: Are you wanting to leave these commented lines

ryanhiebert (approved):
Looks nice!

tractor/_trionics.py
goodboy: nice catch 😏

goodboy:
@ryanhiebert just fixed that last bit and removed all refs to `Actor.statespace`; that should be the last of it 🤞🏼

Issue #175, opened 2020-12-22
Motivating examples for the readme


goodboy:
Clearly, we need some fancier and more immediately practical examples in our readme (says every person who's glanced ever), so I figured I'd start list of what I think would be coolio.

#### A "worker pool" example copying `multiprocessing`'s [prime number calculator](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor-example)
- it is the example for the [`ProcessPoolExecutor`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor) (ugh) so we'll need to probably mock up the [executor api](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor) and show the SC way of doing things

#### A "controlled fork blast" example
- this would be an example of creating a [fork bomb](https://en.wikipedia.org/wiki/Fork_bomb) out of nested nurseries which can be "controlled" via a SIGINT from the user once they've decided that enough CPU has been crushed; this demonstrates very succinctly the true power of SC.
- we might want to monitor system resources for this test to avoid making it impossible to send input whatsoever

#### A real-time streaming sorting algorithm
- I was thinking of implementing an async-ified version of [pythonmaxmin](https://github.com/lemire/pythonmaxmin) which is based on the paper [here](https://arxiv.org/abs/cs/0610046)
- Another thing we might be able to show off is performing this sort on multiple streams in parallel and using a `trio`-ized version of an [Rx `merge`](http://introtorx.com/Content/v1.0.10621.0/12_CombiningSequences.html#Merge) to serialize results.

Lurker opinions always welcome.

parity3:
Wow, some of these, particularly the stream sorting may do well as their own separate repos/libraries!

Another possible example is to implement one of the function-as-a-service protocols with some tuning goodies.

goodboy:
> Another possible example is to implement one of the function-as-a-service protocols with some tuning goodies.

@parity3 sweet idea. I'd welcome a PR for something that too 😉 

> function-as-a-service 

Also check #182, I think the caps based RPC is going to be here to stay the more I've thought about it.
I actually have been dreaming up auto-generating capn proto schema from python's type annotations.

parity3:
Another piece of code that may be nice to see is for a tractor controller app to run as a daemon that some other sync or polyglot app could call into, and be used as a glorified task queue implementation. You could make it compatible with celery or dramatiq or whatever.

goodboy:
[...]

Issue #77, opened 2019-05-07
Do we need to support invoking sync functions?


goodboy:
@jtrakk made a good point [in gitter](https://gitter.im/python-trio/general?at=5cd0f94e3b6cb0686ab737d4) that supporting both async and regular functions for remote invocation is a bit odd and probably unnecessary. I think I agree; there ain't nothing a regular func can do that an async one can't. Unless I'm forgetting something I think it should be fine to just remove support for regular functions?

Off hand it would allow removing the slew of checks in [`_actor._invoke()`](https://github.com/goodboy/tractor/blob/master/tractor/_actor.py#L62).

ryanhiebert:
I'd be in favor of only supporting async functions, for exactly the reasons you describe. I don't think there's anything significant to be gained, but there is additional complexity that could easily be avoided.

goodboy:
See the comment in richardsheridan/trio-parallel#2 which I think is more fuel for this change.

goodboy:
Yah, I'm thinking for a `.alpha0` release we just drop remote sync function calling support and defer to @richardsheridan's project.

goodboy:
Yeah so TLDR: is that if we're calling sync funcs in some remote actor you'll need the `tractor` runtime **anyway** which means you should have the ability to wrap the call in an async func - so there should be no benefit to supporting that wrapping inside [`_actor._invoke`](https://github.com/goodboy/tractor/blob/master/tractor/_actor.py#L73).

I also think we should make a alt-projects list in the readme and point to others for this kind of thing.

goodboy:
We already have a pointer to `trio-parallel` in the readme, so just need a PR removing the support now.

PR #176, opened 2020-12-23
Add our version of the std lib's "worker pool"

goodboy:
This is a draft of the `tractor` way to implement the example from the
"processs pool" in the stdlib's `concurrent.futures` docs:

https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor-example

Our runtime is of course slower to startup but once up, we get the same performance, this confirms that we need to focus some effort on warm up and teardown times.  The mp forkserver method definitely improves startup delay; rolling our own will likely be a good hot spot to play with.

What's really nice is our implementation is done in approx one 10th of the code ;)

Also, do we want offer and interface that yields results as they arrive?

Relates to #175

What y'all think?

Results from running both scripts:

Ours:
```python
(3.9)  >>> python examples/concurrent_actors_primes.py  
112272535095293 is prime: True
112272535095293 is prime: True
115280095190773 is prime: True
112582705942171 is prime: True
1099726899285419 is prime: False
115797848077099 is prime: True
processing took 1.8208930492401123 seconds
script took 2.9907846450805664 seconds
```

stdlib's

```python
3.9)  >>> python examples/concurrent_futures_primes.py
112272535095293 is prime: True
112582705942171 is prime: True
112272535095293 is prime: True
115280095190773 is prime: True
115797848077099 is prime: True
1099726899285419 is prime: False
processing took 1.8223650455474854 seconds
script took 1.845200538635254 seconds
```


examples/concurrent_actors_primes.py
```diff
@@ -0,0 +1,119 @@
...
+
+    async with tractor.open_nursery() as tn:
+
+        portals = []
+        results = []
+
+        for i in range(workers):
+
+            # this starts a new sub-actor (process + trio runtime) and
+            # stores it's "portal" for later use to "submit jobs" (ugh).
```
goodboy, line 61: typo `it's`..

examples/concurrent_actors_primes.py
```diff
@@ -0,0 +1,119 @@
...
+            portals.append(
+                await tn.start_actor(
+                    f'worker_{i}',
+                    rpc_module_paths=[__name__],
```
goodboy, line 65: I was thinking about changing this name to `exposed_modules`.. what y'all think?

examples/concurrent_actors_primes.py
```diff
@@ -0,0 +1,119 @@
...
+                )
+            )
+
+        async def map(
+            worker_func: Callable[[int], bool],
+            sequence: List[int]
+        ) -> List[bool]:
```
goodboy, line 72: gah should be `List[Tuple[int, bool]]` pretty sure

examples/concurrent_actors_primes.py
```diff
@@ -0,0 +1,119 @@
...
+
+            # define an async (local) task to collect results from workers
+            async def collect_portal_result(func, value, portal):
+
+                results.append((value, await portal.run(func, n=value)))
```
goodboy, line 77: This is a use case for keeping sync functions as per #77?

I realize we could just have wrapped `func` here inside an async function wrapper - is the extra overhead something we should help avoid? does it really matter?

examples/concurrent_actors_primes.py
```diff
@@ -0,0 +1,119 @@
...
+                for value, portal in zip(sequence, itertools.cycle(portals)):
+
+                    n.start_soon(
+                        collect_portal_result,
+                        worker_func,
+                        value,
+                        portal
+                    )
+
+            return results
```
goodboy, line 90: We could offer an incremental yielding system here but it will require adding a `trio` mem channel.

Not sure if this will make the example too complicated?

examples/concurrent_futures_primes.py
```diff
@@ -0,0 +1,40 @@
...
+    for i in range(3, sqrt_n + 1, 2):
+        if n % i == 0:
+            return False
+    return True
+
+def main():
+    with concurrent.futures.ProcessPoolExecutor() as executor:
+        start = time.time()
+
+        for number, prime in zip(PRIMES, executor.map(is_prime, PRIMES)):
```
goodboy, line 31: techincally this [is yielding results as they arrive underneath](https://github.com/python/cpython/blob/63298930fb531ba2bb4f23bc3b915dbf1e17e9e1/Lib/concurrent/futures/process.py#L562) so I guess to be fair we should do that too?

goodboy:
Give that new example a shot with 
`$TERM -e watch -n 0.1  "pstree -a $$" & python examples/parallelism/we_are_processes.py || kill $!`

examples/concurrent_futures_primes.py
richardsheridan: I definitely agree with this. It is trivial for a user to reorder their results when necessary from an as-completed stream but impossible to get results as fast as possible when the library is holding them hostage for ordering!

goodboy:
Still TODO:
- [ ] `aclosing()` around the yielded async gen
- [ ] yield a separate (cloned?) recv_chan and a new submit_job_chan or something like that to get `Executor.submit()` style
     > To make this work, snd_chan should be closed when the jobs are exhausted, so each send_result should receive a clone of snd_chan and do the async with snd_chan: await snd_chan.send(...) trick. (from @richardsheridan)

goodboy:
@richardsheridan i think that get's `aclosing()` in.

I'll make a new issue for a more "serious" worker pool recipe to continue the `.submit()` style api yah?

richardsheridan:
I'm a little perplexed between the state of this pr and #163 . It seems you made the `aclosing` changes over there but not over here?

goodboy:
@richardsheridan haven't rebased that PR yet onto this history 😂 

This one is just to get the examples in; that one is to document them in the README.

goodboy:
Yah, `aclosing()` is [right here](https://github.com/goodboy/tractor/pull/176/files#diff-68aacb5f6d6c4bfc96f32d6c413c666acfd477d76000e0e6b2fc93a12c0ff5f3R108).

PR #178, opened 2020-12-25
Denoise logging

goodboy:
Some minimal effort to reduce `error` logging to warnings.
Change default logging level to `error` to aid with this.

### DIFF
--- a/tractor/_actor.py
+++ b/tractor/_actor.py
@@ -25,7 +25,8 @@ from .log import get_logger
 from ._exceptions import (
     pack_error,
     unpack_error,
-    ModuleNotExposed
+    ModuleNotExposed,
+    is_multi_cancelled,
 )
 from . import _debug
 from ._discovery import get_arbiter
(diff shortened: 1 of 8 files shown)


tractor/_actor.py
```diff
@@ -129,22 +130,27 @@ async def _invoke(
 
     except (Exception, trio.MultiError) as err:
 
-        if not isinstance(err, trio.ClosedResourceError):
-            log.exception("Actor crashed:")
+        # TODO: maybe we'll want differnet "levels" of debugging
+        # eventualy such as ('app', 'supervisory', 'runtime') ?
+        if not isinstance(err, trio.ClosedResourceError) and (
```
goodboy, line 135: Pretty sure this won't end up biting us - it's basically only going to enter the debugger when we don't have a broken channel error or a mult-`Cancelled`

tractor/log.py
```diff
@@ -31,11 +31,13 @@
     'GARBAGE': 1,
     'TRACE': 5,
     'PROFILE': 15,
+    'RUNTIME': 500,
```
goodboy, line 34: Yeah, basically this will be the "logging we want users to actually see by default", above 'error' but below 'critical'. It's white and shouldn't be too invasive or alarming.

guilledk (approved):

Issue #177, opened 2020-12-23
Heh, yeah I guess tractor.run() is kinda dumb..


goodboy:
This is somewhat of a 🤦🏼 and directly related to #168.

Thanks to the `trio` crowd for pointing out that we can probably just make ourselves a:

```python

import tractor
import trio

async main():
    async with tractor.open_runtime():
        .. do tractor things since this is a task...
        
trio.run(main)
```

Some benefits to this:
- it's not scary since it's `trio.run()` 🤯 
- we can run certain parts of code with `debug_mode=True` ?
- the whole process tree can be restarted easily from scratch 
- we aren't hijacking `trio.run()` when we really have no reason to

Some worries:
- starting multiple "runtimes" (aka root actors) in a single program will cause mailbox collisions
  - maybe we could have a `.open_root_runtime()` which can only be called once per program?
  - is this solved by just avoiding channel server address collisions?
- maybe more?

Is there a better name?
- `.open_root()`
- `.start_runtime()`
- `start_root()`

Lurkerzzzz!!!

richardsheridan:
IMHO the ideal API would be:

```python
async with tractor.open_actor_nursery() as n:
    n.start_soon(async_fn)

```

And tractor should take care of any global bookkeeping internally, if there is such a thing.  $0.02 :)

goodboy:
@richardsheridan only problem is there might only ever be *one* "actor_nursery* open per program.

Also note, in current parlance, an `ActorNursery` is a nursery [which spawns actors](https://github.com/goodboy/tractor/blob/master/tractor/_trionics.py#L25) - aka known as new processes running `trio.run()`. So I don't think calling the top level entry point for starting all the machinery/runtime that makes actor spawning possible a "nursery" necessarily makes sense.

maybe `.open_root_actor()`?

goodboy:
Hmm thinking more on this I wonder if we can just do implicit runtime startup (only in the root process) on the first entry to `tractor.open_nursery()`?

I'm thinking there should still be an explicit `.start_root_actor_runtime()` or whatever but that can be implicitly called if we find `tractor.is_root_process() == True` when entering the `.open_nursery()` block.

goodboy:
The removal of `tractor.run()` is going to require a big edit of the test set and examples.
- [ ]  [`tractor_test`](https://github.com/goodboy/tractor/blob/master/tractor/testing/_tractor_test.py#L57) is going to need to be redone
[...]

Issue #168, opened 2020-11-24
Starting the runtime from `trio.run()`


goodboy:
Was chit chatting with some `trio` peeps and realized we can probably make it easy to start our runtime from any of
- `trio.run()`
- `trio.Nursery.start_soon()`
- `trio.Nursery.start()`

by simply factoring out our [main entry point](https://github.com/goodboy/tractor/blob/master/tractor/__init__.py#L145) into a public routine.

The only tricky case might be where client code wants to start it from non-`trio` code which in theory we could use some kind of thread spawning interface? Not sure how far to peer down this rabbit hole (yet).

goodboy:
Resolved by #185.

PR #179, opened 2020-12-27
Deprecate `tractor.run()`

goodboy:
Starting work toward dropping `tractor.run()` as it pertains to #168 and #177.

The ideal user facing API instead looks like:

```python
import trio
import tractor

async def main():

    # actor runtime is started implicitly inside the first used nursery if this actor is the root
        portal = await n.run_in_actor(trio.sleep(1))
        
trio.run(main)
```

or, if the user wants to start up the actor runtime / tree manually:

```python
import trio
import tractor

async def main():

    # this starts up the IPC / channel machinery without having to open an actor nursery.
    # you could also just open an actor nursery and sleep in it's body I guess?
    async with tractor.open_root_actor() as actor:
        # run a top level root actor and that's it.
        await trio.sleep_forever()
        
trio.run(main)
```

Would appreciate thoughts @richardsheridan when you get a chance.

### DIFF
--- a/tractor/_root.py
+++ b/tractor/_root.py
@@ -6,6 +6,7 @@ from functools import partial
 import importlib
 from typing import Tuple, Optional, List, Any
 import typing
+import warnings
 
 import trio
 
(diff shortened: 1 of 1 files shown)


goodboy:
Yeah, sorry this one got auto-merged during rebase onto `master`.

PR #181, opened 2020-12-27
Deprecate `tractor.run()`

goodboy:
Starting work toward dropping `tractor.run()` as it pertains to #168 and #177.

The ideal user facing API instead looks like:

```python
import trio
import tractor

async def main():

    # actor runtime is started implicitly inside the first used nursery if this actor is the root
        portal = await n.run_in_actor(trio.sleep(1))
        
trio.run(main)
```

or, if the user wants to start up the actor runtime / tree manually:

```python
import trio
import tractor

async def main():

    # this starts up the IPC / channel machinery without having to open an actor nursery.
    # you could also just open an actor nursery and sleep in it's body I guess?
    async with tractor.open_root_actor() as actor:
        # run a top level root actor and that's it.
        await trio.sleep_forever()
        
trio.run(main)
```

Would appreciate thoughts @richardsheridan when you get a chance.


richardsheridan:
Looks good to me. Have you documented the use-case for starting up the runtime manually anywhere?

goodboy:
Thanks for taking a look 😄 

> Have you documented the use-case for starting up the runtime manually anywhere?

@richardsheridan not yet but that will come once we remove it. This PR is to prep for doing that but without breaking anything to start.

That main use case btw is likely starting a root and then waiting for requests that dynamically spawn subactors to do work; doing that inside an already open nursery will look ugly and be strange to handle for the respawn case likely.

ryanhiebert:
Thanks for thinking of me. I don't think I'm in a good position to approve this pull request, because I don't understand the factors that play into it. I've been following along, and I don't see any issues with the direction its heading, but my knowledge is also limited enough that I recognize my input on this isn't really going to be meaningful, unfortunately.

Issue #183, opened 2020-12-28
Remove ``tractor.run()``


goodboy:
Once #181 lands and all dependent projects are ported, we can safely move to the new `trio.run()` based entry api.

The ideal user facing API instead looks like:

```python
import trio
import tractor

async def main():

    # actor runtime is started implicitly inside the first used nursery if this actor is the root
    async with tractor.open_nursery() as n:
        portal = await n.run_in_actor(trio.sleep(1))
        
trio.run(main)
```

or, if the user wants to start up the actor runtime / tree manually:

```python
import trio
import tractor

async def main():

    # this starts up the IPC / channel machinery without having to open an actor nursery.
    # you could also just open an actor nursery and sleep in it's body I guess?
    async with tractor.open_root_actor() as actor:
        # run a top level root actor and that's it.
        await trio.sleep_forever()
        
trio.run(main)
```

This is of course all up for discussion.
I'm prone to maybe offer a `tractor.lowlevel` (like `trio`) for stuff like this `open_root_actor()`.

Luckily since most of our docs reference examples, one those are changed there should be little work to do on that front.

Issue #182, opened 2020-12-27
Change `.rpc_module_paths` -> `expose_modules`?


goodboy:
Currently the way to *expose* which python modules can be loaded and subsequently executed via RPC requests is determined by the `rpc_module_paths` [kwarg to `ActorNursery.start_actor()`](https://github.com/goodboy/tractor/blob/master/tractor/_trionics.py#L56).

I think `expose_modules` is a better name.

If anyone has any better ideas please chime.

parity3:
If you could tuck the CBS concept into a plugin or middleware-style module and have an alternative or bare-implementation which allows everything to be called by everyone, that would be a nice-to-have for me. For my uses I envision no need for security and/or it will be handled at a different level, ie in the functions themselves and/or via firewalls or third party RBAC like hashicorp vault/boundary, etc.

It would just be one more thing to think about / manage / get in the way for many use cases IMO.

goodboy:
> If you could tuck the CBS concept into a plugin or middleware-style module and have an alternative or bare-implementation which allows everything to be called by everyone, that would be a nice-to-have for me. 

I think this somewhat defeats the point then 😸 
If you want to allow every actor to be able to call everything then simply load `enable_modules=sys.modules.keys()`, however currently this will result in each actor importing all modules at startup.

>It would just be one more thing to think about / manage / get in the way for many use cases IMO.

Imo it should be this way, much like SC forces you to reason about task lifetimes up front.

goodboy:
You'll also find that pretty much every *RPC system* (especially in python) started out in "bare mode" and eventually changes to an explicit capabilities mode with the "bare mode" as an opt in - this is a more robust and safe design.

Eg. with [rpyc "new style"](https://rpyc.readthedocs.io/en/latest/tutorial/tut3.html#part-3-services-and-new-style-rpyc).

parity3:
I'm impartial to whether CBS is the default, but I do recognize that is the way the industry/community leans. I just think sometimes putting security first tends to tax both the documentation and library adoption. I tend to roll my eyes when looking at documentation about a new tool where the first 2 pages are about creating ACLs, generating certs and adding oauth headers, nothing to do with the task at hand (at least in the beginning). This isn't quite so bad in this case because of the decorators and, well, one can work around it with limited overhead with def remote_eval()... or better yet, just monkeypatch out the CBS code.

[...]

PR #185, opened 2021-01-03
Implicit runtime

goodboy:
Another step toward #183 if we want to do it before first dev release.

I'd also be partial to getting in #182 deprecation (or just changing it outright) here.


goodboy:
Bueller?

Issue #145, opened 2020-08-11
Hard kill tests


goodboy:
As per some light commentary in #144, we need some *hard kill* tests for wen processes fail due to internal machinery failure. An easy way to test this now should be to simply cancel one of the actor internal nurseries and ensure we don't get indefinite hangs; containing nursery/supervisor logic should hard kill (via signal) the subprocess after a suitable timeout.

I've seen this work in practice during manual use but we of course want to prevent regressions.

PR #188, opened 2021-01-09
We aint got zombie shields

goodboy:
@guilledk that thing i was talkin bout..

This definitely pertains to #145 but does not yet solve it.

### DIFF
--- a/examples/multiple_streams_one_portal.py
+++ b/examples/multiple_streams_one_portal.py
@@ -1,0 +1,45 @@
+import trio
+import tractor
+
+
+log = tractor.log.get_logger('multiportal')
+
+
+async def stream_data(seed=10):
+    log.info("Starting stream task")
+
+    for i in range(seed):
+        yield i
+        await trio.sleep(0)  # trigger scheduler
+
+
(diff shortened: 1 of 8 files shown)


examples/multiple_streams_one_portal.py
```diff
@@ -0,0 +1,45 @@
+import trio
```
goodboy, line 1: This is just an example that shows the hang in practice.

The problem is if we have a nursery exit before receiving a ctr-c we'll be stuck in the shielded `proc.wait()` and have no way for the user to trigger graceful exit...

tractor/_spawn.py
```diff
@@ -273,10 +273,14 @@ async def new_proc(
...
 
+                # TODO: No idea how we can enforce zombie
+                # reaping more stringently without the shield
+                # we used to have below...
+
                 # always "hard" join sub procs:
                 # no actor zombies allowed
-                with trio.CancelScope(shield=True):
-                    await proc.wait()
+                # with trio.CancelScope(shield=True):
```
goodboy, line 282: The main change.

Thoughts on this very welcome.

PR #189, opened 2021-01-24
Add a @pub kwarg to allow specifying a "startup response message"

goodboy:
Add a way to specify an initial "startup response message" when first connecting to a `@tractor.msg.pub` streaming function.

`@msg.pub` is a provisional and undocumented api for a small internal dynamic pub-sub system.

This will likely need some tests and thought about if it's superfluous.
The main use case we had in `piker` was for some kind of synchronization/handshake for order management daemons connecting to data feeds on spawn.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ with open('docs/README.rst', encoding='utf-8') as f:
 
 setup(
     name="tractor",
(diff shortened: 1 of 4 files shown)


goodboy:
Replaced by #258

PR #190, opened 2021-01-24
Actor state via messages

goodboy:
An example for @fjarri doing "actor state mutation via messages".

Feel free to criticize the heck out of it.

If you want to see how subactors could modify each other's state we can add that as well.


examples/actors/mutate_remote_state.py
```diff
@@ -0,0 +1,51 @@
+from itertools import cycle
+from pprint import pformat
+
+import trio
+import tractor
+
+_snd_chan, _recv_chan = trio.open_memory_channel(100)
```
goodboy, line 7: sorry ignore this..
got left in from another idea.

examples/actors/mutate_remote_state.py
```diff
@@ -0,0 +1,51 @@
...
+_actor_state = {'some_state_stuff': None}
```
goodboy, line 8: This is a module level variable meaning it maintains it's state for the entire process lifetime.
We could have just as easily made some function that "sleep forever" and wakes up incrementally to report it's state if need be.

examples/actors/mutate_remote_state.py
```diff
@@ -0,0 +1,64 @@
+from itertools import cycle
+from pprint import pformat
+from dataclasses import dataclass, field
+
+import trio
+import tractor
+
+
+@dataclass
+class MyProcessStateThing:
```
goodboy, line 10: Instead of this what are you after?

A function that creates some object and then makes that object mutateable from another inbound message?

goodboy: Maybe you want something like `ray`s "actors" (which i would argue aren't really "actor model" actors):
https://docs.ray.io/en/latest/actors.html

We can also accomplish this but it will require  a slight bit more machinery.

goodboy:
How's that 2nd example @fjarri for a class / [ray style](https://docs.ray.io/en/latest/actors.html) actors api?

examples/actors/ray_style_classes.py
```diff
@@ -0,0 +1,153 @@
...
+) -> Any:
+    """Update process-local state from sent message and exit.
+
+    """
+    # Create a new actor instance per call.
+    # We can make this persistent by storing it either
+    # in a global var or are another clas scoped variable?
+    # If you want it somehow persisted in another namespace
+    # I'd be interested to know "where".
+    actor = ActorState()
```
goodboy, line 52: I get this isn't *ideal* (though it really is no performance hit) in an idiomatic python sense, but the alternative is some other way to store this instance across function-task calls. 

The normal way would be a module level variable (since they're "globally scoped") but I guess in theory you could have a function that stays alive and constantly passes the instance to other tasks over a memory channel - still in that case how does the new task get access to the channel handle?). The alternative is a module level class which has a class level variable which is again globally scoped on the class.

examples/actors/ray_style_classes.py
```diff
@@ -0,0 +1,153 @@
...
+        msg: dict,
+    ) -> Any:
+        return await self._portal.run(
+            _run_proxy_method,
+            meth=meth,
+            msg=msg
+        )
+
+
+def get_method_proxy(portal, target=ActorState) -> MethodProxy:
```
goodboy, line 81: In cases anyone gets *cranky* about this, from `pykka` docs:
> The proxy object will use introspection to figure out what public attributes and methods the actor has, and then mirror the full API of the actor. Any attribute or method prefixed with underscore will be ignored, which is the convention for keeping stuff private in Python.

goodboy: From https://www.pykka.org/en/latest/quickstart/#actor-proxies

fjarri:
Sorry, there's been a lot of input from you in different channels, I'll summarize my thoughts here.

I'm afraid I still don't see the functionality (or even ways to implement the functionality) I need. When I think of an actor system, these are some of things I'm worried about:
- how can a child send a message to its parent?
- if a child wants to terminate or restart itself, how will the parent handle it?
- how can I make message handlers async? How can I guarantee that as long as one message handler of an actor is running, another one will not be started?
- how can I start a child in the same thread/other thread/other process with the same API, without changing the child's code?
- how can I delay the message delivery?
- how can I monitor the state of an actor from another actor (not necessarily the parent)?

I can see how my API (in https://gist.github.com/fjarri/803d2611fc8487c7853f17ed4ad1ec10) *allows* one to implement this functionality. The framework will have enough information to handle these problems. Right now `tractor` seems too low-level for all that. I will have no more help from it as I do from `trio` itself, except for more convenient multiprocessing.

goodboy:
@fjarri thanks so much for the writeup 👍🏼.

I'll address these in order:

> how can a child send a message to its paren

```python
async with tractor.wait_for_actor('parent_name') as portal:
    ... do stuff with portal...
```
This is documented only [very briefly](https://tractor.readthedocs.io/en/latest/#service-discovery). Note the "arbiter" stuff there is wrong and going away.

> if a child wants to terminate or restart itself, how will the parent handle it?

However it wants to. Restarts are a naive basic operation the exact same as spawning or running a task through a portal. If you want a special error that the parent expects to handle then you can just raise it and expect it on the parent's side or you could just have the result of the remote task return a restart message of your choosing. I've pushed up an example showing both multiple tasks within a single subactor restarts as well as restarts of multiple subactors:
https://github.com/goodboy/tractor/pull/190/files#diff-3d37ffe6c0c1d6567affe87d7c0edca4c18adb1d5735075084263351e1f642f8

To understand what's going more easily from a terminal you can run it with the following command:
`$TERM -e watch -n 0.1  "pstree -a $$" & python examples/actors/most_basic_supervisor.py  || kill $!`


> how can I start a child in the same thread/other thread/other process with the same API, without changing the child's code?

I honestly have no idea what you're asking here so you'll have to clarify. There is currently not hot-reloading of code support yet mostly because python doesn't have great facilities for this and thus it needs some thinking but we've had lots of discussion.

> how can I delay the message delivery?

Again don't know what you mean. If you want a reactive style [delay](http://reactivex.io/documentation/operators/delay.html) there are already [`trio` libs underway supporting this](https://github.com/andersea/slurry/blob/fced105d5b4ba96eacbfc65419cbeacf4315e1d1/slurry/sections/_buffers.py#L145).

> how can I monitor the state of an actor from another actor (not necessarily the parent)?

As per the 1st bullet, contact the actor by name, get a portal, send it a message, get a response.
There is no built-in "monitoring" system on purpose for now.

examples/actors/most_basic_supervisor.py
```diff
@@ -0,0 +1,86 @@
...
+    async with tractor.open_nursery() as tn:
+
+        p0 = await tn.start_actor('task_restarter', enable_modules=[__name__])
+
+        # Yes, you can do this from multiple tasks on one actor
+        # or mulitple lone tasks in multiple subactors.
+        # We'll show both.
+
+        async with trio.open_nursery() as n:
+            # we'll doe the first as a lone task restart in a daemon actor
```
goodboy, line 52: brutal typo

goodboy: it's clearly 3 tasks that are constantly restarted...

examples/actors/most_basic_supervisor.py
```diff
@@ -0,0 +1,86 @@
...
+            for i in range(4):
+                n.start_soon(respawn_remote_task, p0)
+
+            # Open another nursery that will respawn sub-actors
```
goodboy, line 56: nope x this. no extra nursery required.. originally it had that but we don't need it.

examples/actors/most_basic_supervisor.py
```diff
@@ -0,0 +1,86 @@
...
+
+            # spawn a set of subactors that will signal restart
+            # of the group of processes on each failures
```
goodboy, line 59: still need to add a loop around the nursery to restart everything if we get a certain error raised. this starts getting into a more formal supervisor strategy api that we have yet to design.

goodboy: hmm actualy just had the idea of doing something like:

```python

async with tractor.supervise_with('one_for_one') as srat:
   async with tractor.open_nursery() as tn:
       .. do tractor stuff
```

@guilledk @salotz thots?

goodboy: hmm we might have to plug into the nursery internals a bit more to get granular control on each sub-proc. needs a little more pondering fo sho.

goodboy: oh and common strat names from erlang:
https://learnyousomeerlang.com/supervisors

goodboy: actually i guess you'd want to pass the nursery *into* the strat so that you get the granular control.. hmm that might compose nicely actually. Then in theory you could use a stack to compose multiple strats?

Woo this actually will be interesting i'm thinking.

goodboy:
FYI we can probably vastly improve the example "proxy" code here with the new native bidir streaming support in #209 🏄🏼

examples/actors/ray_style_classes.py
goodboy, line 1: i think this one could be salvaged in some form?

examples/actors/most_basic_supervisor.py
goodboy, line 1: to salvage?

examples/actors/mutate_remote_state.py
goodboy, line 1: worth modernizing and keeping as well?

goodboy:
> follow-up plan authored by `opencode`

### Salvage before closing this PR

- [ ] Rebuild the actor-state lesson from
  `examples/actors/mutate_remote_state.py` and
  `examples/actors/ray_style_classes.py` on current APIs.

- [ ] Prefer one bounded `examples/stateful_actor.py` that demonstrates
  persistent per-process state, updates two actors, retrieves snapshots,
  and asserts actor isolation.

- [ ] Use current `Portal.run()` or `Portal.open_context()` semantics and
  explicit actor teardown. Drop the reflective Ray-style method proxy and
  its accidental class-variable persistence.

- [ ] Decide separately whether `most_basic_supervisor.py` should become a
  finite task-versus-process restart recipe using `Portal.run()` and
  `tractor.to_actor.run()`.

- [ ] Add docs-example smoke coverage and link the stateful example from
  the RPC/context guide before closing #190 as historical prior art.

Issue #193, opened 2021-02-15
Can we just override the global `breakpoint()` built-in?


goodboy:
Got `breakpoint()` working sanely from sync code recently (PR coming) and am realizing maybe we can just avoid any `await tractor.breakpoint()` stuff by doing a bit of sync blocking inside our internal [`_bp()`](https://github.com/goodboy/tractor/blob/master/tractor/_debug.py#L209) closure?

The main outstanding problem is allowing the `trio` scheduler to increment from sync code, which right now seems is only possible from another thread or using some other finagling like [with `greenback`](https://github.com/oremanj/greenback#example) which will unfortunately make the debugged task slower after engagement. 

The simplest single step forward would mean enforcing `await breakpoint()` from async code (and errorring if the returned coroutine is not awaited) and then just keeping the `breakpoint()` support (that's coming) as is.

PR #194, opened 2021-02-21
Support sync code breakpointing via built-in

goodboy:
Override `breakpoint()` for sync code making it work
properly with `trio` as per:

https://github.com/python-trio/trio/issues/1155#issuecomment-742964018

Relates to #193

Pretty sure this also gets us working `asyncio` actor debugging (from manual anectdote) but needs further testing on that branch.


goodboy:
This is what you get 👻 city.

PR #197, opened 2021-02-24
Drop run

goodboy:
This should start resolving the following:
- [ ] #183
- [x] #182
- [x] repair the bug found in #163 
- [x]  #177
  - The removal of `tractor.run()` is going to require a big edit of the test set and examples.
  - [x]  [`tractor_test`](https://github.com/goodboy/tractor/blob/master/tractor/testing/_tractor_test.py#L57) is going to need to be redone
  - [x]  the boatload of [tests still using it](https://github.com/search?q=tractor.run+repo%3Agoodboy%2Ftractor+path%3Atests&type=Code)
  - [x] all the [docs examples](https://github.com/search?q=tractor.run+repo%3Agoodboy%2Ftractor+path%3Aexamples&type=Code&ref=advsearch&l=&l=)
- [x]  #168

PR #198, opened 2021-02-25
Kinda drop run

goodboy:
Brings in the main fix from #197 without the bulk code base changes to drop `tractor.run()` and `rpc_module_paths` since apparently changing the tests results in the `mp` backend hanging 😂 

This should be enough to at least get #163 in.


goodboy:
Lol, let me guess, the change to find `root` broke something..

PR #201, opened 2021-02-27
Single func example

goodboy:
Figured, might as well and give a shout out to @richardsheridan.

First release version is added here.
Possibly `versioneer` usage and a `setup.cfg` coming in follow up PR shortly.


goodboy:
Yeah so probably just gonna do a first `.alpha0` and we'll follow up later with more sophisticated release tools.

This first pkg on pypi has been too long in the tooths already.

Issue #53, opened 2019-01-14
Two way streaming?


tgoodlet:
I've already left some [notes in the code](https://github.com/tgoodlet/tractor/blob/master/tractor/_actor.py#L93) about how we could do two way streaming using the native `received = yield sent` generator semantics but it's probably worth looking at how other projects are approaching it and if it's even a good idea.

Some projects I found through different communities recently:
- bidirectional streaming in google's [gprc](https://grpc.io/docs/tutorials/basic/java.html#bidirectional-streaming-rpc)
- [streaming in `faust`](https://faust.readthedocs.io/en/latest/userguide/streams.html) (though I think this is moreso a demonstration of a need for a proper `asyncitertools`

Another question is how to accomplish this with traditional messaging patterns like those found in `nng`. There was some [suggestions in gitter](https://gitter.im/nanomsg/nanomsg?at=5c32d78a12db9607e74acdc5) about combining protocols/socket types.

More to come...

ryanhiebert:
This is a really interesting idea. I'm not totally sure what to make of it, but it's very interesting. It seems like it would enable some kind of reactive-programming model across machines, which is a trippy thought. Like CycleJS for distributed systems. I'm not sure what the target application would be for such a system, because I'm remarkably uncreative.

goodboy:
The more I think about this the more I think we should be adopting the multi-task style considerations being discussed in the following `trio` related issues:
- https://github.com/python-trio/trio/issues/719
- https://github.com/python-trio/trio/issues/1110
- https://github.com/python-trio/trio/issues/987
- https://github.com/python-trio/trio/issues/959

The entire "channels" related issue list is also worth keeping and eye on:
https://github.com/python-trio/trio/issues?q=is%3Aopen+channels

We already have this api to run an async func and have it stream but currently there's no way to have the caller send values to the calee. In essence we want the same streaming behavior as if you had to actors call each other and receive stream results but in a more compact *single-context-channel* thing API.

The thought i had was something like the following based on our [context api](https://tractor.readthedocs.io/en/latest/#channels-and-contexts):

```python
@tractor.stream
async def streamer(
   ctx: tractor.Context,
) -> None:
   """A simple web response streaming server.
   """
      tx, rx = context.attach_stream() 
      while True:
      
          url = await rx.recv()
          val = await web_request(url)

          # this is the same as ``yield`` in the async gen case
          await tx.send(val)


async def caller(urls: List[str]):
   async with tractor.open_nursery() as n:
[...]

PR #203, opened 2021-03-02
Add support for stream cloning

goodboy:
Leverage `trio`'s built-in rx channel cloning with our `ReceiveStream` to enable multi-consumer usage.

### DIFF
--- a/tractor/_portal.py
+++ b/tractor/_portal.py
@@ -71,15 +71,18 @@ class ReceiveStream(trio.abc.ReceiveChannel):
         try:
             msg = await self._rx_chan.receive()
             return msg['yield']
+
         except trio.ClosedResourceError:
             # when the send is closed we assume the stream has
             # terminated and signal this local iterator to stop
             await self.aclose()
             raise StopAsyncIteration
+
         except trio.Cancelled:
             # relay cancels to the remote task
             await self.aclose()
             raise
+
         except KeyError:
             # internal error should never get here
             assert msg.get('cid'), (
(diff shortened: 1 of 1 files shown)


tractor/_portal.py
```diff
@@ -102,14 +105,25 @@ async def aclose(self):
...
+            # if we've been cloned don't kill the stream
+            log.debug("there are still consumers running keeping stream alive")
+            return
+
         if self._shielded:
             log.warning(f"{self} is shielded, portal channel being kept alive")
             return
 
+        # close the local mem chan
+        rx_chan.close()
```
goodboy, line 125: This likely needs to be put above.

Thanks to @richardsheridan for catching 🏄🏼

goodboy:
Heh, so this doesn't even work remotely as planned 😂 

Turns out receive mem chans [do not support broadcasting in `trio`](https://trio.readthedocs.io/en/stable/reference-core.html#trio.MemoryReceiveChannel):
> The clones all share the same underlying channel. Whenever a clone receive()s a value, it is removed from the channel **and the other clones do not receive that value**. If you want to send multiple copies of the same stream of values to multiple destinations, like itertools.tee(), then you need to find some other solution; this method does not do that.

This actually ended up being a source of major confusion in pikers/piker#158 where this PR's change to `ReceiveStream.aclose()` was actually preventing far end producer cancellation and then a very hard to debug back-pressure instigated lockup 😂 

So basically I think we need to drop this idea wholesale (at least if we're sticking with *clone* as it has meaning in `trio`) and instead venture into what we're actually after: **broadcast channels** as in python-trio/trio#987.

goodboy:
See also #53 for further discussion on how we don't want `trio`'s `.clone()` semantics (since it isn't useful cross-process).

PR #205, opened 2021-04-27
Drop sync func invocation support.

### LINKED ISSUE #77: Do we need to support invoking sync functions?

goodboy:
Resolves #77.

Realized we had a bunch of arbiter methods that were sync and of course now the concurrent primes example function needs to be async. Looks a little unwelcoming imo but then again if peeps are here and don't want to make everything `async` maybe it's the wrong spot 😂 

Also, as mentioned in the readme, [`trio-parallel`](https://github.com/richardsheridan/trio-parallel) for anyone who wants the sync func support.

### DIFF
--- a/examples/actor_spawning_and_causality.py
+++ b/examples/actor_spawning_and_causality.py
@@ -1,7 +1,7 @@
 import tractor
 
 
-def cellar_door():
+async def cellar_door():
     assert not tractor.is_root_process()
     return "Dang that's beautiful"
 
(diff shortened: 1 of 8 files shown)


goodboy:
@richardsheridan if there's any other pointers you'd want us to have to your repo just say the word.

Issue #172, opened 2020-12-18
Do we need a streaming equivalent (sugar) for `.run_in_actor()`?


goodboy:
So in other words something like `ActorNursery.stream_from_actor()` which can be used to explicitly call a streaming function (with checks for types built-in) which would avoid the future-y feeling way you'd do it right now with the returned portal.

An example in our [test set here](https://github.com/goodboy/tractor/pull/171/files#diff-d960bb61fde148c9fabda1375f3cb82be40ccc75ac1260cb4353d26cfcedb5d1R277) looks like this:

```python
async with tractor.open_nursery() as n:

    stream = await(await n.run_in_actor(
        'streamer',
        stream_data,
        seed=11,
    )).result()
```

Which we could maybe switch to this?

```python
async with tractor.open_nursery() as n:

        stream = await n.stream_from_actor(
            stream_data,
            seed=11,
        )
```

This does add another method to the API but it is also more explicit and allows for type checking the remote routine at spawn time. Some other notes:
- it's the same number of `await`s as the `.run_in_actor()` usage which 
- doing an `async for portal.result():` is kind of confusing and unintuitive since in the streaming case the portal returns effectively an `AsyncIterator` "proxy" of sorts to allow client code to consume values from IPC - in the `Portal.run()` case (when using `.start_actor()`) it seems less strange since you're more explicitly expecting back the value you'd get if you called the streaming function locally.
- we might want to think about how cross-actor context manager APIs might look want to look for "one-off" calls as well since this seems like a useful api addition
- maybe we should scrap all these methods and go to a `trio` nursery style `.start()` and `.start_soon()`? Currently we're always using a `.start()` style where the parent waits for the child to come up before unblocking - #155

Opinions very welcome 🤔

goodboy:
#287 and #290 are related to this and would likely deprecate and remove this.

PR #206, opened 2021-04-28
Stream contexts

goodboy:
This is a breaking change to remove support for [streaming using `Portal.run()`](https://tractor.readthedocs.io/en/latest/#asynchronous-generators) calls and instead replace it with a new `@asynccontextmanager` api `Portal.open_stream_from() as stream:`.

The main motivation for this is that it avoids leaving users to do there own manual stream shutdown management (normally ala `async_generator.aslosing()` or the new `aclosing()` coming in the stdlib) and promotes more stringent SC style in streaming clients. That latter part is important as we move shortly to supporting an api for #53 which will require explicit *cross-actor-context* management.

The basic gist should be clear from the changed tests.
Once this is merged **it will break old code**.

### DIFF
--- a/tests/test_discovery.py
+++ b/tests/test_discovery.py
@@ -268,28 +268,30 @@ async def close_chans_before_nursery(
                 portal2 = await tn.start_actor(
                     'consumer2', enable_modules=[__name__])
 
-                async with (
-                    portal1.open_stream_from(stream_forever) as agen1,
-                    portal2.open_stream_from(stream_forever) as agen2,
-                ):
-                    async with trio.open_nursery() as n:
-                        n.start_soon(streamer, agen1)
-                        n.start_soon(cancel, use_signal, .5)
-                        try:
-                            await streamer(agen2)
-                        finally:
-                            # Kill the root nursery thus resulting in
-                            # normal arbiter channel ops to fail during
-                            # teardown. It doesn't seem like this is
-                            # reliably triggered by an external SIGINT.
-                            # tractor.current_actor()._root_nursery.cancel_scope.cancel()
+                # TODO: compact this back as was in last commit once
+                # 3.9+, see https://github.com/goodboy/tractor/issues/207
+                async with portal1.open_stream_from(stream_forever) as agen1:
+                    async with portal2.open_stream_from(
+                        stream_forever
+                    ) as agen2:
+                        async with trio.open_nursery() as n:
+                            n.start_soon(streamer, agen1)
+                            n.start_soon(cancel, use_signal, .5)
+                            try:
+                                await streamer(agen2)
+                            finally:
+                                # Kill the root nursery thus resulting in
+                                # normal arbiter channel ops to fail during
+                                # teardown. It doesn't seem like this is
+                                # reliably triggered by an external SIGINT.
+                                # tractor.current_actor()._root_nursery.cancel_scope.cancel()
 
-                            # XXX: THIS IS THE KEY THING that happens
-                            # **before** exiting the actor nursery block
+                                # XXX: THIS IS THE KEY THING that happens
+                                # **before** exiting the actor nursery block
 
-                            # also kill off channels cuz why not
-                            await agen1.aclose()
-                            await agen2.aclose()
+                                # also kill off channels cuz why not
+                                await agen1.aclose()
+                                await agen2.aclose()
         finally:
             with trio.CancelScope(shield=True):
                 await trio.sleep(1)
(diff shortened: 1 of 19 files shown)


guilledk (approved):
New API looking sexy! :tada:

ryanhiebert (approved):
That sounds like a great motivation. I'm not good for a deep review, but nice!

tests/test_streaming.py
```diff
@@ -272,48 +278,55 @@ async def test_respawn_consumer_task(
...
-                    # respawn
-                    cs = await ln.start(consume)
+                    if not expect:
+                        print("all values streamed, BREAKING")
+                        break
 
-                if not expect:
-                    print("all values streamed, BREAKING")
-                    break
+        # TODO: this is justification for a
```
goodboy, line 330: Ahh so one question is if we want to move forward with a `ActorNursery.run_in_actor()` equivalent for streams.  Pretty sure we go over this in https://github.com/goodboy/tractor/issues/172.

tractor/_streaming.py
```diff
@@ -18,32 +21,242 @@ class Context:
...
+        'stream' not in params
+    ):
         raise TypeError(
             "The first argument to the stream function "
             f"{func.__name__} must be `ctx: tractor.Context`"
         )
     return func
+
+
+class ReceiveMsgStream(trio.abc.ReceiveChannel):
```
goodboy, line 122: Stole this name from @njsmith 🤫

tractor/_streaming.py
```diff
@@ -18,32 +21,242 @@ class Context:
...
+    def __init__(
+        self,
+        ctx: Context,
+        rx_chan: trio.abc.ReceiveChannel,
+        portal: 'Portal',  # type: ignore # noqa
+    ) -> None:
+        self._ctx = ctx
+        self._rx_chan = rx_chan
+        self._portal = portal
+        # self._chan = portal.channel
```
goodboy, line 144: clean this?

tractor/_streaming.py
```diff
@@ -18,32 +21,242 @@ class Context:
...
+
+    # delegate directly to underlying mem channel
+    def receive_nowait(self):
+        return self._rx_chan.receive_nowait()
+
+    async def receive(self):
+        try:
+            msg = await self._rx_chan.receive()
+            return msg['yield']
+            # return msg['yield']
```
goodboy, line 155: and this.

tractor/_portal.py
```diff
@@ -39,113 +41,30 @@ async def maybe_open_nursery(
...
+    )
 
 
 class Portal:
     """A 'portal' to a(n) (remote) ``Actor``.
 
-    Allows for invoking remote routines and receiving results through an
-    underlying ``tractor.Channel`` as though the remote (async)
-    function / generator was invoked locally.
+    A portal is "opened" (and eventually closed) by one side of an
```
goodboy, line 54: Hopefully this explanation is more acceptable 🏄🏼

tractor/_portal.py
```diff
@@ -369,6 +221,122 @@ async def cancel_actor(self):
...
+                    f'{func} must be a non-streaming async function!')
+
+            fn_mod_path, fn_name = func_deats(func)
+
+        return await self._return_once(
+            *(await self._submit(fn_mod_path, fn_name, kwargs))
+        )
+
+    @asynccontextmanager
+    async def open_stream_from(
```
goodboy, line 288: This is the meat of the new API.

tests/test_discovery.py
```diff
@@ -260,36 +263,36 @@ async def close_chans_before_nursery(
...
-
-                        # also kill off channels cuz why not
-                        await agen1.aclose()
-                        await agen2.aclose()
+                portal1 = await tn.start_actor(
+                    name='consumer1', enable_modules=[__name__])
+                portal2 = await tn.start_actor(
+                    'consumer2', enable_modules=[__name__])
+
+                async with (
```
goodboy, line 271: heh, yeah so turns out this style (tuple after the with) is not allowed pre-3.9 😢; though oddly [doesn't seem documented anywhere](https://docs.python.org/3/whatsnew/3.9.html)?

It's def failing [in ci](https://github.com/goodboy/tractor/runs/2459473341?check_suite_focus=true#step:5:35) where you can see there's no syntax error in 3.9.

So we'll have to delay this I guess until we do a 3.9+ pin.

goodboy: Made a wish list: https://github.com/goodboy/tractor/issues/207

goodboy:
Oh right, docs heh.

I guess that's a thing if there are any not already in the examples?

Issue #134, opened 2020-07-27
Should we just discard multiprocessing?


goodboy:
Now that #128 is landing we don't really have a dire need for the stdlib's `multiprocessing` especially since it has proven much less reliable then pure `trio` alternatives (which I guess just means async wrapping around `subprocessing`?).

Reasons to keep it for now:
- windows peeps might find they need it (though CI seems to say the native `trio` spawner is doing just fine)
- benchmarking against the `forkserver` will probably be handy when we get to rolling our own for linux
- it's probably worth doing another comb through of the code to be sure we aren't missing anything critical for our own spawning stuff

Reasons to dump:
- it's super annoying to test reliably in CI
- it pickles state and sends to the child
- it's currently be hacked to heck in [`_forkserver_override.py`](https://github.com/goodboy/tractor/blob/master/tractor/_forkserver_override.py) and if we're unlucky will have to be again with Python 3.9..
- you can't use the new debugging stuff from #129 with it

Relates to #84 and of course always open to opinions.

PR #208, opened 2021-04-29
Mp teardown hardening

goodboy:
Might solve some of the complaints in any of:
#84 #89 #134 #146

Still want to test enabling the nested spawning tests again to see if we get any improvements.

### DIFF
--- a/tractor/_spawn.py
+++ b/tractor/_spawn.py
@@ -283,14 +283,15 @@ async def new_proc(
                 await proc.wait()
 
             log.debug(f"Joined {proc}")
-            # pop child entry to indicate we are no longer managing this subactor
+            # pop child entry to indicate we no longer managing this subactor
             subactor, proc, portal = actor_nursery._children.pop(subactor.uid)
 
             # cancel result waiter that may have been spawned in
             # tandem if not done already
             if cancel_scope:
                 log.warning(
-                    f"Cancelling existing result waiter task for {subactor.uid}")
+                    "Cancelling existing result waiter task for "
+                    f"{subactor.uid}")
                 cancel_scope.cancel()
     else:
         # `multiprocessing`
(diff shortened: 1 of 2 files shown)


tractor/_spawn.py
```diff
@@ -356,23 +404,53 @@ async def new_proc(
...
+                    # in that case, wait shortly then kill the process.
+                    reaping_cancelled = True
+
+                    if proc.is_alive():
+                        with trio.move_on_after(0.1) as cs:
+                            cs.shield = True
+                            await proc_waiter(proc)
+
+                        if cs.cancelled_caught:
+                            proc.terminate()
```
richardsheridan, line 438: I think it is necessary to end this block with an unconditional `raise`. Although most cases would work, I suppose some edge cases would break down the line. For example, consider if you needed to directly cancel the top-level nursery of this function, and then check it's `cancelled_caught` property (similar to the `move_on_after` in this block).  This exception handler would discard the cancel and it would appear to never catch any cancels.

goodboy: Thanks for detailed explanation @richardsheridan, will definitely add!

tractor/_spawn.py
```diff
@@ -356,23 +404,53 @@ async def new_proc(
...
+
+            if not reaping_cancelled:
+                if proc.is_alive():
```
goodboy, line 441: So maybe we can just rely on the proc being dead here?
I guess it would make more sense to poll and then check?

If that works we can probably drop the flag set and check.

goodboy:
@richardsheridan this is more acceptable yah?

tractor/_spawn.py
goodboy: The necessary bit is tagged on the last logic block.

Issue #204, opened 2021-04-13
We need broadcast channels, stat.


goodboy:
This ties in with #40 and #53 both of which heavily rely on the discussion in https://github.com/python-trio/trio/issues/987.

The main gotcha is that `trio`'s mem chans don't have broadcasting built-in (frankly making them mostly an SC anti-pattern if you have followed the `.clone()` semantics discussion). This mis-assumption was what originally fueled #203.

The obvious main use case is for broadcasting stream data on the *receive* side of a stream such that multiple rx-side tasks get each their own copy of a new message. NB: We already [have a prototype](https://github.com/goodboy/tractor/blob/master/tractor/msg.py#L96) for producer side ala `tractor.msg.pub`, though I'm not sure it's the best solution either.

Depending on how we move forward with bidirectional-streaming in #53 it might be worth offering a `.clone()` or similar method which provides a broadcast reference token to consumer tasks.

Definitely needs a deeper dive.

For reference here's the current short list of broadcast chan impls:
- the [`eventual_values`](https://trio-util.readthedocs.io/en/v0.5.0/#trio_util.AsyncValue.eventual_values) predicate matching iterator in `trio-util`
- the "lossless", slowest consumer limiting, event queue [`aioevent` gist](https://gist.github.com/basak/007da3fc2448300d037c9bb008cc5e80) from @basak 
- [`slurry`'s `Pipeline.tap()`](https://github.com/andersea/slurry/blob/master/slurry/_pipeline.py#L68) effectively does the,
  > Apply backpressure based on the slowest consumer:
  from https://github.com/python-trio/trio/issues/987#issue-425510847
  
  but, with non-zero buffer size reccomendations:
  >  max_buffer_size: Although not recommended in general, it is possible to
            set a buffer on the output channel. (default ``0``) See

Any more implementations are greatly welcome from the lurker pack 😉

goodboy:
Interesting stuff from the `rust` method njs linked: https://tokio-rs.github.io/tokio/doc/tokio/sync/broadcast/index.html#lagging

> This broadcast channel implementation handles this case by setting a hard upper bound on the number of values the channel may retain at any given time. This upper bound is passed to the channel function as an argument.

> If a value is sent when the channel is at capacity, the oldest value currently held by the channel is released. This frees up space for the new value. Any receiver that has not yet seen the released value will return RecvError::Lagged the next time recv is called.

[...]

Issue #36, opened 2018-09-01
Spec out the async-ipc protocol


tgoodlet:
`tractor` utilizes a simple multiplexed protocol for conducting inter-process-task-communication (IPTC)?

Each per-process `trio` task can invoke tasks in other processes and received responses depending on the type of the remote *callable*. All packets are encoded as `msgpack` serialized dictionaries which I'll refer to as *messages*.

How it works
-----------------
When an actor wants to invoke a remote routine it sends a `cmd` packet:
`{'cmd': (ns, func,  kwargs,  uid, cid)}` of type `Dict[str, Tuple[str, str, Dict[str, Any], Tuple[str, str], str]]`
Where:
- `ns` is the remote module name
- `func` is the remote function name
- `kwargs` is a `dict` of keyword arguments to call the function with
- `uid` is the unique id of the *calling* actor
- `cid` is the unique id of the *call* by a specific task

The first response is a *function type* notifier msg:
`{'functype': functype, 'cid': cid}` of type `Dict[str, str]`. 
Where `functype` can take one of:
- `'function'` for a plain ol' function
- `'asyncfunction'` for an asynchronous function
- `'asyncgen'` for an asynchronous generator

Depending on the value of `functype` then the following message(s) are sent back to the caller:
- `'function'`, `'asyncfunction'`:
   - a single packet with the remote routine's result `{'return', result, 'cid', cid}` of type `Dict[str, Any]`
- `'asyncgen'`:
   - a stream of packets with the remote async generator's sequence of results `{'yield', value, 'cid', cid}` of type `Dict[str, Any]`.

What should be done
-----------------------------
- spec out this protocol and define it somewhere in the code base with proper [generic type definitions](https://mypy.readthedocs.io/en/latest/generics.html) that can be used to type check related internal ipc apis (see @vodik's [comment](https://github.com/tgoodlet/tractor/pull/35#discussion_r214507776) in #35).
- work towards full *bi-directional* async generator support i.e. [using `.asend()`](https://www.python.org/dev/peps/pep-0525/#asynchronous-generator-object) also mentioned in an internal [todo](https://github.com/tgoodlet/tractor/blob/master/tractor/_actor.py#L96).
[...]

PR #209, opened 2021-05-02
Bidir streaming

goodboy:
First draft bidirectional streaming support as discussed in #53.

Further todos:
- [x] more extensive tests for closing down either side of the stream early
- [x] port `_debug.py` remote tty locking to context api.
- [ ] possibly a better internal design for tracking bidir vs unidir context usage to avoid hacky `if not self._ctx._portal` checking inside `ReceiveMsgStream.aclose()`
- [x] tests where the consumer tasks use `async for stream` while sender is running independently in another trio task
- [ ] docs on the new apis
- [ ] should we add broadcasting support here for #204 ?
- [ ] move to a split `SendMsgStream` / `ReceiveMsgStream` type set and staple them together using a channel/messaging equivalent of [`trio.StapledStream`](https://trio.readthedocs.io/en/stable/reference-io.html#trio.StapledStream)?
  - currently this would allow making the `stream` arg to `@stream` funcs a `SendMsgStream` instead of a `Context`
  - turns out [`anyio` has something similar: `StapledObjectStream`](https://anyio.readthedocs.io/en/stable/api.html#anyio.streams.stapled.StapledObjectStream) (obvs we won't use `*Object*` since we're `msgpack` type contrained). Actual code is [here](https://github.com/agronholm/anyio/blob/master/src/anyio/streams/stapled.py#L46).
- [ ] example in the main readme. i'm thinking that should be a big boon when compared to projects like [`faust`](https://faust.readthedocs.io/en/latest/userguide/streams.html) and [`ray`](https://docs.ray.io/en/master/async_api.html) which (i think) are unidirectional only?
- [ ] from https://github.com/pikers/piker/pull/190
  > would be nice if in tractor we can require either a `ctx` arg, or a named arg with `ctx` in it **and** a type annotation of `tractor.Context` instead of strictly requiring a ctx arg.

Critique and suggestions very welcome from lurkers 😎.

### DIFF
--- a/tests/test_2way.py
+++ b/tests/test_2way.py
@@ -52,9 +52,14 @@ async def assert_state(value: bool):
     assert _state == value
 
 
-def test_simple_contex():
+@pytest.mark.parametrize(
+    'error_parent',
+    [False, True],
+)
+def test_simple_context(error_parent):
 
     async def main():
+
         async with tractor.open_nursery() as n:
 
             portal = await n.start_actor(
@@ -74,10 +79,19 @@ def test_simple_contex():
             # after cancellation
             await portal.run(assert_state, value=False)
 
+            if error_parent:
+                raise ValueError
+
             # shut down daemon
             await portal.cancel_actor()
 
-    trio.run(main)
+    if error_parent:
+        try:
+            trio.run(main)
+        except ValueError:
+            pass
+    else:
+        trio.run(main)
 
 
 @tractor.context
--- a/tests/test_2way.py
+++ b/tests/test_2way.py
@@ -1,0 +1,140 @@
+"""
+Bidirectional streaming and context API.
+"""
+
+import trio
+import tractor
+
+# from conftest import tractor_test
+
+# TODO: test endofchannel semantics / cancellation / error cases:
+# 3 possible outcomes:
+# - normal termination: far end relays a stop message with
+# final value as in async gen from ``return <val>``.
+
+# possible outcomes:
+# - normal termination: far end returns
+# - premature close: far end relays a stop message to tear down stream
+# - cancel: far end raises `ContextCancelled`
+
+# future possible outcomes
+# - restart request: far end raises `ContextRestart`
+
+
+_state: bool = False
+
+
+@tractor.context
+async def simple_setup_teardown(
+
+    ctx: tractor.Context,
+    data: int,
+
+) -> None:
+
+    # startup phase
+    global _state
+    _state = True
+
+    # signal to parent that we're up
+    await ctx.started(data + 1)
+
+    try:
+        # block until cancelled
+        await trio.sleep_forever()
+    finally:
+        _state = False
+
+
+async def assert_state(value: bool):
+    global _state
+    assert _state == value
+
+
+def test_simple_contex():
+
+    async def main():
+        async with tractor.open_nursery() as n:
+
+            portal = await n.start_actor(
+                'simple_context',
+                enable_modules=[__name__],
+            )
+
+            async with portal.open_context(
+                simple_setup_teardown,
+                data=10,
+            ) as (ctx, sent):
+
+                assert sent == 11
+
+                await portal.run(assert_state, value=True)
+
+            # after cancellation
+            await portal.run(assert_state, value=False)
+
+            # shut down daemon
+            await portal.cancel_actor()
+
+    trio.run(main)
+
+
+@tractor.context
+async def simple_rpc(
+
+    ctx: tractor.Context,
+    data: int,
+
+) -> None:
+
+    # signal to parent that we're up
+    await ctx.started(data + 1)
+
+    print('opening stream in callee')
+    async with ctx.open_stream() as stream:
+
+        count = 0
+        while True:
+            try:
+                await stream.receive() == 'ping'
+            except trio.EndOfChannel:
+                assert count == 10
+                break
+            else:
+                print('pong')
+                await stream.send('pong')
+                count += 1
+
+
+def test_simple_rpc():
+    """The simplest request response pattern.
+
+    """
+    async def main():
+        async with tractor.open_nursery() as n:
+
+            portal = await n.start_actor(
+                'rpc_server',
+                enable_modules=[__name__],
+            )
+
+            async with portal.open_context(
+                simple_rpc,
+                data=10,
+            ) as (ctx, sent):
+
+                assert sent == 11
+
+                async with ctx.open_stream() as stream:
+
+                    for _ in range(10):
+
+                        print('ping')
+                        await stream.send('ping')
+                        assert await stream.receive() == 'pong'
+
+                # stream should terminate here
+
+            await portal.cancel_actor()
+
+    trio.run(main)
--- a/tests/test_2way.py
+++ b/tests/test_2way.py
@@ -262,7 +262,7 @@ async def test_caller_closes_ctx_after_callee_opens_stream(
             async with ctx.open_stream() as stream:
                 async for msg in stream:
                     pass
-        except trio.ClosedResourceError:
+        except tractor.ContextCancelled:
             pass
         else:
             assert 0, "Should have received closed resource error?"
--- a/tests/test_2way.py
+++ b/tests/test_2way.py
@@ -6,20 +6,13 @@ import pytest
 import trio
 import tractor
 
-# from conftest import tractor_test
+from conftest import tractor_test
 
-# TODO: test endofchannel semantics / cancellation / error cases:
-# 3 possible outcomes:
-# - normal termination: far end relays a stop message with
-# final value as in async gen from ``return <val>``.
-
-# possible outcomes:
-# - normal termination: far end returns
-# - premature close: far end relays a stop message to tear down stream
-# - cancel: far end raises `ContextCancelled`
-
-# future possible outcomes
-# - restart request: far end raises `ContextRestart`
+# the general stream semantics are
+# - normal termination: far end relays a stop message which
+# terminates an ongoing ``MsgStream`` iteration
+# - cancel termination: context is cancelled on either side cancelling
+#  the "linked" inter-actor task context
 
 
 _state: bool = False
@@ -30,6 +23,7 @@ async def simple_setup_teardown(
 
     ctx: tractor.Context,
     data: int,
+    block_forever: bool = False,
 
 ) -> None:
 
@@ -41,8 +35,11 @@ async def simple_setup_teardown(
     await ctx.started(data + 1)
 
     try:
-        # block until cancelled
-        await trio.sleep_forever()
+        if block_forever:
+            # block until cancelled
+            await trio.sleep_forever()
+        else:
+            return 'yo'
     finally:
         _state = False
 
@@ -56,7 +53,14 @@ async def assert_state(value: bool):
     'error_parent',
     [False, True],
 )
-def test_simple_context(error_parent):
+@pytest.mark.parametrize(
+    'callee_blocks_forever',
+    [False, True],
+)
+def test_simple_context(
+    error_parent,
+    callee_blocks_forever,
+):
 
     async def main():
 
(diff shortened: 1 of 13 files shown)


tests/test_2way.py
```diff
@@ -0,0 +1,140 @@
...
+# from conftest import tractor_test
+
+# TODO: test endofchannel semantics / cancellation / error cases:
+# 3 possible outcomes:
+# - normal termination: far end relays a stop message with
+# final value as in async gen from ``return <val>``.
+
+# possible outcomes:
+# - normal termination: far end returns
+# - premature close: far end relays a stop message to tear down stream
```
ryanhiebert, line 17: Near end sends a message, not far end, right?

goodboy: It should eventually be one of either, we'll need tests for both cases.

You should be able to close either side's `Context.open_stream()` and the other side shuts down gracefully i'm thinking?

ryanhiebert: That makes sense to me. I figured you were laying out how it might happen from either side

tractor/_streaming.py
```diff
@@ -247,14 +193,207 @@ async def aclose(self):
...
+            with trio.move_on_after(0.5) as cs:
+                cs.shield = True
+                log.warning(
+                    f"Cancelling stream {cid} to "
+                    f"{self._portal.channel.uid}")
+
+                # NOTE: we're telling the far end actor to cancel a task
+                # corresponding to *this actor*. The far end local channel
+                # instance is passed to `Actor._cancel_task()` implicitly.
+                await self._portal.run_from_ns('self', '_cancel_task', cid=cid)
```
goodboy, line 272: one outstanding question i still have is whether instead of calling this *special cancel method* on the "actor runtime" we should have a special cancel *message*. I mention this same question in #36.

tractor/_streaming.py
```diff
@@ -247,14 +193,207 @@ async def aclose(self):
...
+            shield=shield,
+        ) as rchan:
+
+            if self._portal:
+                self._portal._streams.add(rchan)
+
+            try:
+                yield rchan
+
+            except trio.EndOfChannel:
```
goodboy, line 331: So if we receive a `stop` msg then we don't send one since we can assume the far end has already closed.

tractor/_streaming.py
```diff
@@ -247,14 +193,207 @@ async def aclose(self):
...
+                raise
+
+            else:
```
goodboy, line 334: If we're closing just send the `stop` msg and move on right?

tests/test_advanced_streaming.py
```diff
@@ -0,0 +1,142 @@
...
+
+"""
+import itertools
+from typing import Set, Dict, List
+
+import trio
+import tractor
+
+
+_registry: Dict[str, Set[tractor.ReceiveMsgStream]] = {
```
goodboy, line 12: Linking comment https://github.com/goodboy/tractor/commit/f9dd2add43a478eeb9a6f8a8d852c23953553e5f#r50700925.

goodboy: @gc-ss mentioned,

> … but if you don't like that, what if we make _registry an argument to publisher?

Yup can do that, I just didn't bc turns out the consumers can update via sending updates on their individual streams.

> It would be perfect if publisher accepted an environment variable (or a Context?)

This can also be done though I'm not sure what you mean by *environment*. A `Context` isn't really required explicitly here but could be used to control cancellation if wanted.

For an alt to `_registry` we could do something like,

```python
class PublisherState:
    subscriptions: dict[str, tractor.MsgStream] = {}
```
kinda thing?

tests/test_advanced_streaming.py
```diff
@@ -0,0 +1,142 @@
...
+                itertools.cycle(_registry.keys())
+            ):
+                await n.run_in_actor(
+                    consumer,
+                    name=f'consumer_{sub}',
+                    subs=[sub],
+                )
+
+            # make one dynamic subscriber
+            await n.run_in_actor(
```
goodboy, line 129: Hmm, I'm thinking we could just move the dynamic case into the immediate task here instead of sleeping?

tractor/_streaming.py
```diff
@@ -247,14 +193,207 @@ async def aclose(self):
...
+                # signal ``StopAsyncIteration`` on far end.
+                await self.send_stop()
+
+            finally:
+                if self._portal:
+                    self._portal._streams.remove(rchan)
+
+    async def started(self, value: Any) -> None:
```
goodboy, line 342: Hmm, I'm thinking we should support bare calling: `.started()` [just like `trio` allows with `TaskStatus.started()`](https://github.com/python-trio/trio/blob/master/trio/_core/_run.py#L730)?

tests/test_advanced_streaming.py
gc-ss: > I'm not sure what you mean by environment. A Context isn't really required explicitly here but could be used to control cancellation if wanted

By environment or Context, I meant the context/environment/configuration due to which the function might behave in a non-deterministic manner but the values of which don't necessarily change between function invocations - like a remote host address:port or username/password.

In this case - think about two actors communicating with each other. The exact same actors (types) might behave in very different ways if their context/environment/configuration were different with everything else (eg: arguments) being the same.

Unlike arguments though, their context/environment/configuration dont change between function invocations (like arguments do).

For example, the registry does not need to change between function invocations.

tests/test_advanced_streaming.py
```diff
@@ -0,0 +1,142 @@
...
+                consumer,
+                name='consumer_dynamic',
+                subs=list(_registry.keys()),
+            )
+
+            # block until cancelled by user
```
gc-ss, line 135: block until cancelled by user in 10 seconds?

tractor/_actor.py
```diff
@@ -404,17 +444,23 @@ async def _push_result(
...
+        #     # indicate to consumer that far end has stopped
+        #     return await send_chan.aclose()
 
         try:
             log.debug(f"Delivering {msg} from {actorid} to caller {cid}")
             # maintain backpressure
             await send_chan.send(msg)
 
         except trio.BrokenResourceError:
+            # TODO: what is the right way to handle the case where the
```
gc-ss, line 458: This is a very difficult decision to make. It would have been less complex with old school unidirectional streams.

We might look into using both NACKs and ACKs. This coupled with atleast once delivery can ensure durable message processing

tests/test_advanced_streaming.py
goodboy: I mean, yeah a service actor can easily create wtv objects they need and mutate them as actor-local state.
They can go further and offer a seperate api to mutate that object/state from other actors if needed.

I'm just not sure we need to show that in *this* example since the point is just a super basic dynamic pubsub system.
We do have a [pretty terrible example](https://tractor.readthedocs.io/en/latest/#actor-local-aka-process-global-variables) in the docs that could be improved.

Bonus points for sweet PRs 😉

tractor/_actor.py
goodboy: > This coupled with atleast once delivery can ensure durable message processing

Not sure what you mean.
TCP already has reliability built in, for other protocols this might be some for a user to consider yes.

Also, the messaging *processing* isn't the problem here, it's whether a stop/cancel message made it to the far end *and* what to do in specifically the case where no ack for that msg is received.
That's just [two general's](https://en.wikipedia.org/wiki/Two_Generals'_Problem#Engineering_approaches).

I'm pretty sure (at least right now) `trio` takes the same naive approach of just "hoping for the best".

goodboy: For interested lurkers:

> For example, the first general could send 100 messengers, anticipating that the probability of all being captured is low. With this approach the first general will attack no matter what, and the second general will attack if any message is received. Alternatively the first general could send a stream of messages and the second general could send acknowledgments to each, with each general feeling more comfortable with every message received. As seen in the proof, however, neither can be certain that the attack will be coordinated.

TCP obvs already has built-in mechanisms for it's reliability requirements so currently we don't have to think too hard since failures should bubble up from the transport layer.

The main question was moreso about cancellation race conditions that can arise where the local channel is killed after it's sent the stop and whether or not we should wait / shield the mem chan until the msg is processed later (also presumably this is all before the sub-actor is killed).

goodboy: Also to clarify, `tractor`'s streams can in theory be restarted but i'm pretty sure internal feeder mem chans should be re-created?

tractor/_actor.py
```diff
@@ -404,17 +444,23 @@ async def _push_result(
         send_chan, recv_chan = self._cids2qs[(actorid, cid)]
         assert send_chan.cid == cid  # type: ignore
 
-        if 'stop' in msg:
-            log.debug(f"{send_chan} was terminated at remote end")
-            # indicate to consumer that far end has stopped
-            return await send_chan.aclose()
+        # if 'stop' in msg:
```
goodboy, line 447: yeah i think we can drop this now.
processing is now inside the `ReceiveMsgStream.receive()` body.

tractor/_debug.py
```diff
@@ -38,7 +37,9 @@
 _in_debug = False
 
 # lock in root actor preventing multi-access to local tty
-_debug_lock = trio.StrictFIFOLock()
+_debug_lock: trio.StrictFIFOLock = trio.StrictFIFOLock()
```
goodboy, line 47: Might pull this debugger stuff into a new PR to keep things more separate?

The main thing was being able to use the context api for tty locking which is a lot cleaner now 🏄🏼

goodboy: Nah, keeping it here for now; it's all internals anyway.

tests/test_advanced_streaming.py
```diff
@@ -0,0 +1,142 @@
...
+@tractor.context
+async def subscribe(
+
+    ctx: tractor.Context,
+
+) -> None:
+
+    global _registry
+
+    # syn caller
```
goodboy, line 50: `s/syn/sync`

goodboy:
There's still quite a few things on the todo list but I think this is good enough to get peeps using on `master` and we can always follow up.

tractor/_streaming.py
goodboy: This is in now.

tests/test_advanced_streaming.py
goodboy: Changed this to be shorter for CI / real testing.

goodboy:
Lol i'm stressing here about what the heck is wrong with CI... we have a 10m timeout in the config 😂

goodboy:
grr windows..

goodboy:
Yeah, so not really sure what's going on with CI here.

I can't get the 3.9 run to fail on that debugger test and windows is just hanging in general..

It seems if i do one thing to try and emulate old stream tear down behavior for windows then linux will break, and vice versa 😂 

Logically the semantics of the code make sense to me as of 569426b we might need a windows tester to see if they can dig in ...

ping @chrizzFTD

goodboy:
Just pushed up the `windows_bi_streaming` branch: https://github.com/goodboy/tractor/tree/windows_bi_streaming to see if it's the new tests causing the hangs..

CI run is [here](https://github.com/goodboy/tractor/actions/runs/940848516).

goodboy:
Aight, so looks to be [the multi-task test that beats up windows..](https://github.com/goodboy/tractor/actions/runs/940886662).

My guess is race conditions again ... smh.

goodboy:
Huhhh.

Well good news is that new work on the `transport_hardening` branch seems to have resolved the windows CI hangs/failures.

So likely that means either the actual transport error handling tweaks or the new fully working *root can't clobber child in tty lock* work is somehow avoiding some other side effect.

Technically the context api semantics were further refined, so could also be that.

Either way that works for me.

Next thing is to factor those commits back to this branch from the dev history off of `infect_asyncio` and make sure everything still works 😂

goodboy:
Ok so I've put up #219 as a replacement for this which pulls out all debugger work that relies on this change set.
This will hopefully make the patch simpler to grok and also more easy to resolve wtv strange CI stuff is going on.

I will be moving the todo list to that PR as well.

goodboy:
Replaced by #219.

PR #212, opened 2021-06-02
Try msgspec

goodboy:
First draft trying out `msgspec` over `msgpack-python`.

~We currently can't easily decode streams since there's no api support~, see:
https://github.com/jcrist/msgspec/issues/27, see the updates; we just need to implement our own framing with a bytes len prefix 🥳 

The bulk of speed improvements are supposed to come from decoding + validation so we're not really seeing benefits from either of those yet 😂 

TODO:
- [ ] add nested structs for our ipc protocol for #36 
- [ ] offer a way to specify which serialization lib is used at runtime startup (likely much like the `start_method` kwarg to `open_root_actor()`), maybe `serialization_method`?
- [ ] fix up [the remaining failing tests which are shipping `defaultdict`s](https://github.com/goodboy/tractor/pull/212/checks?check_run_id=2806536791#step:5:1232) for some reason
- [ ] appease `mypy` stuff
- [ ] potentially offer `msgpack-numpy` as the optional dep if we end up liking the performance here?
- [ ] oh right, pin to python 3.8+ (since `msgspec` requires it) and we might as well sine we're onto a new alpha release anyway
- [ ] document `msgspec` use in the readme

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,13 +38,20 @@ setup(
         'tractor.testing',
     ],
     install_requires=[
+
+        # trio related
         'trio>0.8',
-        'msgpack',
         'async_generator',
+        'tricycle',
+        'trio_typing',
+
         'colorlog',
         'wrapt',
-        'trio_typing',
         'pdbpp',
+
+        # serialization
+        'msgpack',
+        'msgspec',
     ],
     tests_require=['pytest'],
     python_requires=">=3.7",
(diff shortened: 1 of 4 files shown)


goodboy:
Gah next hassle thing is not supporting native `tuple` serialization.
Gonna open another issue.

goodboy:
Dropping this in favor of #214 which branches from master and includes a simpler git history.

PR #214, opened 2021-07-01
Optional msgspec support

goodboy:
Supersedes #212 but adding `msgspec` as an optional dependency which is loaded as the transport backend `TCPMsgspecStream` if importable.

This also reworks TCP transport EOF handling to raise a new internal error type `TransportClosed` which now handled around the core message loop and inter-actor handshake.

This also add `tricycle` as a dep since it has a lot useful stuff anyway that we'll probably use and once we go full `msgspec` after dropping `<= 3.8` python, it'll be necessary.


goodboy:
So turns out those last CI runs never had `msgspec` installed 😂 

This demonstrates 2 interesting outcomes
- `msgspec` hasn't been tested against our test suite (which, shocker, it's completely broken for a variety of reasons)
- apparently the master branch is already seeing problems on windows and some linux runs?
  - this is particularly odd because if you look at the runs the tests complete but the jobs are cancelled?
  - this means the CI failures we're seeing as part of #209 might be false negatives?
  
Lots more work to do here but quick summary is I'm going to defer this support and pull out the transport stuff that's needed for the bi-dir streaming changes and put it as a separate smaller PR onto mainline.

goodboy:
Welp, now all the non-`msgspec` tests are borked 😂 

Will have to dig into it.

PR #217, opened 2021-07-05
Debugger hardening

goodboy:
This solves the *root-clobbers-child-in-tty-lock* issues once and for all 🤞🏼 

I'm also interested to see if this solves the CI issues in #209.

PR #218, opened 2021-07-05
Always hard kill sub-procs on teardown

goodboy:
Trying to see which parts from #217 fix CI in #209.


goodboy:
Clearly didn't solve anything...

Issue #216, opened 2021-07-05
Multi-root discovery: pragmatic, simple consensus.


goodboy:
This issue is a more practical and immediate equivalent of the discussion in #184.

From the start of the project a hacky naive approach to finding the addresses of other "actors" has been this idea of an [arbiter](https://tractor.readthedocs.io/en/latest/#service-discovery) which we [implement as a "special actor"](https://github.com/goodboy/tractor/blob/master/tractor/_actor.py#L1025) with methods for looking up (socket) addresses by name. This is of course not an ideal system since there will always be a race during a multi-tree startup for the "arbiter" address as well no flexible consensus system for how that position can be transferred to another tree / root actor when the first is torn down / fails. The fragility is further emphasized in how root actors "check" for the registry (arbiter) existing which is simply do a fast [TCP connect and drop](https://github.com/goodboy/tractor/blob/master/tractor/_root.py#L108) on the supposed arbiter socket address.

Summarizing the current naive/questionable design for an address registry:
- a single socket address is allocated to some root actor designated the "arbiter" (aka a registry actor) and this address is passed to other python programs which would like to search for actors also using this same registry
- the way to "check" if the arbiter "exists" is to do a nasty TCP connect/drop which results in us having to specially handle and remap `trio.BrokenResourceError`s to an internal `TransportClosed` error which is ignore silently
- there is no mechanism for fail-over, arbiter re-election, transfer of the registry between trees

---
#### Digging into "why" this is in the code:
This "arbiter" idea was originally adopted from other "actor system" projects:
- examples

Places to start some research
- gossip protocol
- matrix "federation replication" api
- raft
- 

WIP, will come back.

goodboy:
resource dump, will do writeup later.

https://en.wikipedia.org/wiki/Fallacies_of_distributed_computing
  https://spec.matrix.org/unstable/server-server-api/#room-state-resolution
  https://spec.matrix.org/unstable/rooms/v2/
  https://github.com/goodboy/tractor/issues/216
  https://github.com/goodboy/tractor/issues/184
  https://en.wikipedia.org/wiki/Multicast_DNS
  https://www.geeksforgeeks.org/difference-between-network-address-translation-nat-and-port-address-translation-pat/?ref=rp
  https://en.wikipedia.org/wiki/Name_Service_Switch
  https://en.wikipedia.org/wiki/Link-Local_Multicast_Name_Resolution
[...]

PR #219, opened 2021-07-05
Bi streaming no debugger stuff

goodboy:
This is a first draft of a *bidirectional streaming* api as discussed in #53.

This specific PR is a clone of the original #209 but with debugger changes factored out to see if we can get a clean CI run as well as a smaller patch set for reviewers. There is a follow up PR #220 which includes the debugger improvements that now use and rely on this new api.

Critique and suggestions very welcome from lurkers 😎.

---
#### continued task list from #209:

- [x] more extensive tests for closing down either side of the stream early
- [x] port `_debug.py` remote tty locking to context api.
- [x] tests where the consumer tasks use `async for stream` while sender is running independently in another trio task
- ~[ ] should we add broadcasting support here for #204 ?~ let's keep it separate.
/api.html#anyio.streams.stapled.StapledObjectStream) (obvs we won't use `*Object*` since we're `msgpack` type contrained). Actual code is [here](https://github.com/agronholm/anyio/blob/master/src/anyio/streams/stapled.py#L46).
- [x] example in the main readme. i'm thinking that should be a big boon when compared to projects like [`faust`](https://faust.readthedocs.io/en/latest/userguide/streams.html) and [`ray`](https://docs.ray.io/en/master/async_api.html) which (i think) are unidirectional only?

#### the following were deferred to #223
- docs on the new apis
- possibly a better internal design for tracking bidir vs unidir context usage to avoid hacky `if not self._ctx._portal` checking inside `ReceiveMsgStream.aclose()`
- from https://github.com/pikers/piker/pull/190
  > would be nice if in tractor we can require either a `ctx` arg, or a named arg with `ctx` in it **and** a type annotation of `tractor.Context` instead of strictly requiring a ctx arg.
- move to a split `SendMsgStream` / `ReceiveMsgStream` type set and staple them together using a channel/messaging equivalent of [`trio.StapledStream`](https://trio.readthedocs.io/en/stable/reference-io.html#trio.StapledStream)? I'm thinking we'll make all streams a `MsgStream` and just don't allow send on receive only (which should be minority use case i'd imagine eventually).
  - currently this would allow making the `stream` arg to `@stream` funcs a `SendMsgStream` instead of a `Context`
  - turns out [`anyio` has something similar: `StapledObjectStream`](https://anyio.readthedocs.io/en/stable
  - - [ ] a comment from [some discussion](https://github.com/goodboy/tractor/pull/209#discussion_r631449147) there:
- 
  >The main question was moreso about cancellation race conditions that can arise where the local channel is killed after it's sent the stop and whether or not we should wait / shield the mem chan until the msg is processed later (also presumably this is all before the sub-actor is killed).
  - this really has to do with whether or not we want a channel *teardown transaction* eventually. I personally think right now it's not a requirement (and makes facing 2-general's more immediate). the more sane thing to address first is the naive discovery issues we have as per #216.

---
#### outstanding windows hangs CI work, leak-over from #209 
- ~[ ] [comment from #209](https://github.com/goodboy/tractor/pull/209#issuecomment-861879386) suggests it might have been the mutlti-task producer task, but removing that on 83dd8e4 shows it's still not working...~ it was some wacky `logging` thing 🙄

### DIFF
--- a/tractor/_actor.py
+++ b/tractor/_actor.py
@@ -64,7 +64,7 @@ async def _invoke(
     tb = None
 
     cancel_scope = trio.CancelScope()
-    cs: trio.CancelScope = None
+    cs: Optional[trio.CancelScope] = None
 
     ctx = Context(chan, cid)
     context: bool = False
--- a/tractor/_actor.py
+++ b/tractor/_actor.py
@@ -46,6 +46,7 @@ class ActorFailure(Exception):
 
 
 async def _invoke(
+
     actor: 'Actor',
     cid: str,
     chan: Channel,
@@ -58,10 +59,12 @@ async def _invoke(
     """Invoke local func and deliver result(s) over provided channel.
     """
     treat_as_gen = False
-    cs = None
+
     cancel_scope = trio.CancelScope()
-    ctx = Context(chan, cid, _cancel_scope=cancel_scope)
-    context = False
+    cs: trio.CancelScope = None
+
+    ctx = Context(chan, cid)
+    context: bool = False
 
     if getattr(func, '_tractor_stream_function', False):
         # handle decorated ``@tractor.stream`` async functions
@@ -149,14 +152,22 @@ async def _invoke(
             # context func with support for bi-dir streaming
             await chan.send({'functype': 'context', 'cid': cid})
 
-            with cancel_scope as cs:
+            async with trio.open_nursery() as scope_nursery:
+                ctx._scope_nursery = scope_nursery
+                cs = scope_nursery.cancel_scope
                 task_status.started(cs)
                 await chan.send({'return': await coro, 'cid': cid})
 
             if cs.cancelled_caught:
+                if ctx._cancel_called:
+                    msg = f'{func.__name__} cancelled itself',
+
+                else:
+                    msg = f'{func.__name__} was remotely cancelled',
+
                 # task-contex was cancelled so relay to the cancel to caller
                 raise ContextCancelled(
-                    f'{func.__name__} cancelled itself',
+                    msg,
                     suberror_type=trio.Cancelled,
                 )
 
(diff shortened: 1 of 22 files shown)


goodboy:
@chrizzFTD if you've got a second I would sure love to know why these tests hang 😂

goodboy:
Looking at [this last run](https://github.com/goodboy/tractor/pull/219/checks?check_run_id=2998743088) it appears as though it's the discovery tests causing the hang now?

I think we need a manual tester to come in and check it out.

goodboy:
Ok going to increment in from #220 onto the `CI_increment_for_windows_bidirstreaming` branch to see if we can bisect the solution 😂

goodboy:
As per whatever dark magic is going on [in this patch set](https://github.com/goodboy/tractor/pull/222/checks?check_run_id=3012040900) this should solve windows CI hangs?

goodboy:
Lol true 🤯 right there.

goodboy:
Aight, gonna consider this a clean run since the one flaky test is just in-determinism that'll be removed in #220 😎 

Last couple things I'm hoping to get in here:
- readme examples of streaming
- possibly allow passing in a nursery to `.open_context()`
  - might need a couple tests around this
- traceback truncating to remove all our internal frames

tractor/_actor.py
```diff
@@ -378,10 +473,24 @@ async def _stream_handler(
         try:
             await self._process_messages(chan)
         finally:
+
+            # channel cleanup sequence
```
goodboy, line 477: Ahh dang I guess we can drop this?

tractor/_actor.py
```diff
@@ -411,22 +528,32 @@ async def _push_result(
...
+        # actorid = chan.uid
+        assert chan.uid, f"`chan.uid` can't be {chan.uid}"
+        send_chan, recv_chan = self._cids2qs[(chan.uid, cid)]
         assert send_chan.cid == cid  # type: ignore
 
-        if 'stop' in msg:
-            log.debug(f"{send_chan} was terminated at remote end")
-            # indicate to consumer that far end has stopped
-            return await send_chan.aclose()
+        # if 'error' in msg:
```
goodboy, line 536: And this?

tractor/_actor.py
goodboy: Oh wait, no this is a reminder comment of sorts.

PR #220, opened 2021-07-05
Ctx debugger

goodboy:
Clone of #217 but with commits factored to pair with #219.

Again trying to find the source commits that breaks windows CI...

In theory this should have the same CI results as #217 but who knows any more 😂

### DIFF
--- a/examples/debugging/fast_error_in_root_after_spawn.py
+++ b/examples/debugging/fast_error_in_root_after_spawn.py
@@ -1,0 +1,53 @@
+'''
+fast fail test with a context.
+ensure the partially initialized sub-actor process
+doesn't cause a hang on error/cancel of the parent
+nrusery.
+
+'''
+import trio
+import tractor
+
+
+@tractor.context
+async def sleep(
+    ctx: tractor.Context,
+):
+    await trio.sleep(0.5)
+    await ctx.started()
+    await trio.sleep_forever()
+
+
+async def open_ctx(
+    n: tractor._trionics.ActorNursery
+):
+
+    # spawn both actors
+    portal = await n.start_actor(
+        name='sleeper',
+        enable_modules=[__name__],
+    )
+
+    async with portal.open_context(
+        sleep,
+    ) as (ctx, first):
+        assert first is None
+
+
+async def main():
+
+    async with tractor.open_nursery(
+        debug_mode=True,
+        loglevel='runtime',
+    ) as an:
+
+        async with trio.open_nursery() as n:
+            n.start_soon(open_ctx, an)
+
+            await trio.sleep(0.2)
+            await trio.sleep(0.1)
+            assert 0
+
+
+if __name__ == '__main__':
+    trio.run(main)
--- a/examples/debugging/fast_error_in_root_after_spawn.py
+++ b/examples/debugging/fast_error_in_root_after_spawn.py
@@ -1,8 +1,9 @@
 '''
-fast fail test with a context.
-ensure the partially initialized sub-actor process
+Fast fail test with a context.
+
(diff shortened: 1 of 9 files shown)


goodboy:
Ok so same results as #217 so we're approaching some kind of sanity.

This somehow means some patch from the bidir streaming, which is causing the windows run to hang, is somehow fixed by some debugger related patch that uses the new context/bidir streaming apis?

This is extremely strange since the debugger nor the tests for it should not be getting triggered at all on windows 🤯 

Gonna start spelunking in diffs again and likely put up a diff branch to basically `git bisect` where the fix comes in.

goodboy:
Yeah I honestly feel like I'm losing my mind..

- windows doesn't run any debug tests and thus shouldn't trigger the debugger, ever
- there's no code here that changes anything substantial except debugger related logic
- the CI run is working on windows and completing even though it should be agnostic to this patch set

-> The CI without this patch set hangs indefinitely.

This is like the *problem of evil* for testing..

examples/debugging/fast_error_in_root_after_spawn.py
```diff
@@ -0,0 +1,53 @@
+'''
+fast fail test with a context.
+ensure the partially initialized sub-actor process
+doesn't cause a hang on error/cancel of the parent
+nrusery.
```
goodboy, line 5: typo nursery.

tests/test_debugger.py
```diff
@@ -317,32 +317,58 @@ def test_multi_daemon_subactors(spawn, loglevel):
...
+        next_msg = bp_forever_msg
 
     else:
         raise ValueError("Neither log msg was found !?")
 
-    child.sendline('c')
+    # NOTE: previously since we did not have clobber prevention
+    # in the root actor this final resume could result in the debugger
+    # tearing down since both child actors would be cancelled and it was
+    # unlikely that `bp_forever` would re-acquire the tty loack again.
```
goodboy, line 328: typo loack

examples/debugging/root_timeout_while_child_crashed.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    async with tractor.open_nursery(
+        debug_mode=True,
+        loglevel='debug'
+    ) as n:
+
+        # spawn both actors
+        portal = await n.run_in_actor(key_error)
+
+        # XXX: originally a bug causes by this
```
goodboy, line 23: typos causes

tests/test_debugger.py
```diff
@@ -397,7 +423,7 @@ def test_multi_nested_subactors_error_through_nurseries(spawn):
     child = spawn('multi_nested_subactors_error_up_through_nurseries')
 
     # startup time can be iffy
-    time.sleep(1)
+    # time.sleep(1)
```
goodboy, line 426: drop this?

tractor/_debug.py
```diff
@@ -115,21 +123,74 @@ def set_quit(self):
...
 
 @asynccontextmanager
-async def _acquire_debug_lock(uid: Tuple[str, str]) -> AsyncIterator[None]:
-    """Acquire a actor local FIFO lock meant to mutex entry to a local
-    debugger entry point to avoid tty clobbering by multiple processes.
-    """
+async def _acquire_debug_lock(
+    uid: Tuple[str, str]
+) -> AsyncIterator[trio.StrictFIFOLock]:
+    '''Acquire a actor local FIFO lock meant to mutex entry to a local
```
goodboy, line 129: s/to/for

goodboy: maybe use *breakpoint* instead of *entry point*

PR #221, opened 2021-07-06
Ci increment for windows bidirstreaming

goodboy:
Bisecting commits to figure out where/why windows breaks for #219 compared to #220.


goodboy:
If we get to parity with #220 and windows is still hanging i'm going throw in the towel.

goodboy:
Ok so apparently 5ffacab solves the hang?

I guess next thing is to try just putting that logging change stuff on a branch?

goodboy:
Isolated the issue to a weird `logging` related in thing #219...

PR #226, opened 2021-08-01
Debugger tests determinism

goodboy:
Trying to get debugger tests more deterministic.

- Better early timeout handling
- send `continue`s on child re-locks of tty

More coming.

### DIFF
--- a/tractor/_root.py
+++ b/tractor/_root.py
@@ -17,6 +17,7 @@ from . import _spawn
 from . import _state
 from . import log
 from ._ipc import _connect_chan
+from ._exceptions import is_multi_cancelled
 
 
 # set at startup and after forks
@@ -177,7 +178,7 @@ async def open_root_actor(
 
                 entered = await _debug._maybe_enter_pm(err)
 
-                if not entered:
+                if not entered and not is_multi_cancelled(err):
                     logger.exception("Root actor crashed:")
 
                 # always re-raise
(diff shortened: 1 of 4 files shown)


goodboy:
Hmm, yeah so i've seen this 2way test flake a few times:
https://github.com/goodboy/tractor/pull/226/checks?check_run_id=3215310077

It seems to be a race with when teardown of the context happens versus when a stream message arrives. We might want to set some state after a feed mem chan is removed from the msg loop and then not raise these sorts of errors during the termination sequence.

tractor/_root.py
```diff
@@ -177,7 +178,7 @@ async def open_root_actor(
 
                 entered = await _debug._maybe_enter_pm(err)
 
-                if not entered:
+                if not entered and not is_multi_cancelled(err):
```
goodboy, line 181: This avoids stupid long cancelled inceptions in logging.

tractor/_trionics.py
```diff
@@ -281,7 +281,6 @@ async def _open_and_supervise_one_cancels_all_nursery(
                     # Instead try to wait for pdb to be released before
                     # tearing down.
                     if is_root_process():
-                        log.exception(f"we're root with {err}")
```
goodboy, line 284: This slipped through in `.alhpa1`; was originally for debugging the root clobbering child-in-tty stuff..

guilledk (approved):

PR #228, opened 2021-08-03
Add macos-latest to CI's os matrix

guilledk:
This simply adds `macos-latest` to the testing envoirments.

There seems to be a few different failures as well as hangs using the `trio` backend.

You can see the actions run on my fork: https://github.com/guilledk/tractor/actions/runs/1094209479


goodboy:
Nice.

Huh, so somewhat expected the hangs start during cancellation tests (our most aggressive of the bunch).

```
tests/test_cancellation.py FF.F.....F......F.
```

@guilledk would you mind also putting a 5m timeout on the actions run?
I don't think I've seen a clean job that takes longer then that.

goodboy:
Yeah looks like we're gonna need a mac peep to check these out.
I'm frankly somewhat surprised this happens on the `trio`/`subprocess` spawner of all places.

guilledk:
@goodboy Timeout done!

goodboy:
@guilledk lol can we get a new version?

PR #229, opened 2021-08-08
`tokio`-style broadcast channels

goodboy:
~Literally the most prototype thing ever~ actually decently well tested now, but based on discussion in:
- https://github.com/python-trio/trio/issues/987 particularly the [comment on what `rust` has](https://github.com/python-trio/trio/issues/987#issuecomment-817917147)
- https://github.com/goodboy/tractor/issues/204

Nothing more then a runnable module atm..

Some interesting notes:
- to consumers i'm pretty sure this looks *exactly* like a ring buffer
- we should be able to swap in any queue type that supports writing to it (like `deque.appendleft()`) where the oldest value is just discarded, for some examples:
  - [`ringbuf`'s handling of overflow is much the same](https://github.com/elijahr/ringbuf#handling-overflow--underflow)
  - a new [rust based `ringbuf` impl](https://github.com/Supportfactory/ringbuf/blob/master/ringbuf.py#L107)
 
Update:
----------
before i lose this thought...

If we wanted to also support *slowest consumer* style, I'm pretty sure there's an easy hack to just inherit and proxy through reads to `trio._channel.MemoryChannelState.data` (presuming the `queue` size to `broadcast_receiver()` and the underlying mem chan's `deque` are the same size).

The thinking is, there's a length check to determine if the sender should block in `MemorySendChannel.send_nowait()`:
https://github.com/python-trio/trio/blob/master/trio/_channel.py#L147

If `.data` here proxied to `lambda: max(self._subs.values())` and the *order* of these branches were reversed, it might *just work* no? 

```python
        elif len(self._state.data) >= self._state.max_buffer_size:
            # causes ``.send()`` to block
            raise trio.WouldBlock
        else:
            self._state.data.append(value)
```

Follow up:
-------------
- dropped all the *clones* stuff, bleh; a big source of errors and confusion imo
- probably should just use `id(self)` for `.key` instead of `uuid4()`?
- [x] probably still need a couple tests for trying to (re)open a subscription from a now closed caster 
- [x] need a test for the case where `._recv()` is cancelled before the first ask in receives a value and that ensure that no other event queued tasks miss that value and/or re-get their last pulled value
- [x] make an issue detailing the *slowest-consumer* style notes from above. -> #235


tractor/_live_from_tokio.py
```diff
@@ -0,0 +1,194 @@
...
+        self,
+        rx_chan: MemoryReceiveChannel,
+        buffer_size: int = 100,
+
+    ) -> None:
+
+        self._rx = rx_chan
+        self._len = buffer_size
+        self._queue = deque(maxlen=buffer_size)
+        self._subs = {id(current_task()): -1}
```
goodboy, line 44: Apparently we can just hash the `Task` directly?

tractor/_live_from_tokio.py
```diff
@@ -0,0 +1,194 @@
...
+        task = current_task()
+
+        # check that task does not already have a value it can receive
+        # immediately and/or that it has lagged.
+        key = id(task)
+        # print(key)
+        try:
+            seq = self._subs[key]
+        except KeyError:
+            self._subs.pop(key)
```
goodboy, line 59: woops this is wrong.

tractor/_live_from_tokio.py
```diff
@@ -0,0 +1,194 @@
...
+
+            # items with lower indices are "newer"
+            self._queue.appendleft(value)
+
+            # broadcast new value to all subscribers by increasing
+            # all sequence numbers that will point in the queue to
+            # their latest available value.
+            for sub_key, seq in self._subs.items():
+
+                if key == sub_key:
```
goodboy, line 92: maybe do a `._subs` copy and pop instead ?

tractor/_live_from_tokio.py
```diff
@@ -0,0 +1,194 @@
...
+                # #     # already lagged
+                # #     seq = Lagged
+
+                self._subs[sub_key] += 1
+
+            self._value_received = None
+            event.set()
+            return value
+
+        else:
```
goodboy, line 108: comment here that we're a queued task waiting on the latest value, **but not** the first of such tasks.

tractor/_live_from_tokio.py
```diff
@@ -0,0 +1,194 @@
...
+
+    ) -> BroadcastReceiver:
+        key = id(current_task())
+        self._subs[key] = -1
+        try:
+            yield self
+        finally:
+            self._subs.pop(key)
+
+    async def aclose(self) -> None:
```
goodboy, line 130: ye not sure what to do here?
Nothin?

tractor/_live_from_tokio.py
```diff
@@ -0,0 +1,204 @@
...
+
+    async def receive(self):
+
+        task: Task
+        task = current_task()
+
+        # check that task does not already have a value it can receive
+        # immediately and/or that it has lagged.
+        key = id(task)
+        # print(key)
```
goodboy, line 57: clean this.

goodboy:
There. I think this is ready to start adding tests.

I'll leave it for a couple hours to see if any of you lurkers complain 😂

tractor/_broadcast.py
```diff
@@ -0,0 +1,201 @@
...
+    fastest consumer.
+
+    Additional consumer tasks can receive all produced values by registering
+    with ``.subscribe()``.
+
+    '''
+    def __init__(
+        self,
+
+        rx_chan: MemoryReceiveChannel,
```
goodboy, line 38: We should probably switch this to a `tri.abc.ReceiveChannel` right?

Since you should be able to pass in anything that duck-types that interface?

tractor/_broadcast.py
```diff
@@ -0,0 +1,238 @@
...
+                self._queue,
+            )
+        finally:
+            # drop from subscribers and close
+            self._subs.pop(clone)
+            # XXX: this is the reason this function is async: the
+            # ``AsyncResource`` api.
+            await clone.aclose()
+
+    # TODO:
```
goodboy, line 153: Last set of design questions i *think*?

tractor/_broadcast.py
```diff
@@ -0,0 +1,238 @@
...
+                # adhere to ``tokio`` style "lagging":
+                # "Once RecvError::Lagged is returned, the lagging
+                # receiver's position is updated to the oldest value
+                # contained by the channel. The next call to recv will
+                # return this value."
+                # https://tokio-rs.github.io/tokio/doc/tokio/sync/broadcast/index.html#lagging
+
+                # decrement to the last value and expect
+                # consumer to either handle the ``Lagged`` and come back
+                # or bail out on it's own (thus un-subscribing)
```
goodboy, line 81: oof, not `it's`..

tractor/_broadcast.py
```diff
@@ -0,0 +1,238 @@
...
+
+            # items with lower indices are "newer"
+            self._queue.appendleft(value)
+
+            # broadcast new value to all subscribers by increasing
+            # all sequence numbers that will point in the queue to
+            # their latest available value.
+
+            subs = self._subs.copy()
+            # don't decerement the sequence # for this task since we
```
goodboy, line 106: *decrement*

tractor/_broadcast.py
```diff
@@ -0,0 +1,238 @@
...
+            # already retreived the last value
+            subs.pop(key)
+            for sub_key, seq in subs.items():
```
goodboy, line 109: is there some fancier-faster way to do something like this?

tractor/_broadcast.py
```diff
@@ -0,0 +1,238 @@
...
+            self._value_received = None
+            event.set()
+            return value
+
+        # This task is all caught up and ready to receive the latest
+        # value, so queue sched it on the internal event.
+        else:
+            await self._value_received.wait()
+
+            seq = self._subs[key]
```
goodboy, line 122: can we drop all of these sequence # checks?

goodboy:
lol, yeah so since moving to the "clones" version the actual N copies consistency was broken obviously because `._value_received` event wasn't common to all clones 😂 

Fixed this and now am adding a "broadcast state" type much like default mem chans.

goodboy:
Yah so binance feeds proved that the assumption of slept consumers waking up can expect to read the `0` index is definitely false 😂 

We can maybe try this down the road if we invert the indexing?
I gotta think about it.

tractor/_streaming.py
```diff
@@ -253,6 +256,56 @@ async def aclose(self):
...
+        This operation is indempotent and and mutates this stream's
+        receive machinery to copy and window-length-store each received
+        value from the far end via the internally created broudcast
+        receiver wrapper.
+
+        '''
+        if self._broadcaster is None:
+            self._broadcaster = broadcast_receiver(
+                self,
+                self._rx_chan._state.max_buffer_size,
```
goodboy, line 289: This presumes a `trio.MemoryReceiveChannel` which is not ideal, so what would be the best default here and how would we reflect it in typing?

tractor/_streaming.py
```diff
@@ -253,6 +256,56 @@ async def aclose(self):
...
+            )
+            # override the original stream instance's receive to
+            # delegate to the broadcaster receive such that
+            # new subscribers will be copied received values
+            # XXX: this operation is indempotent and non-reversible,
+            # so be sure you can deal with any (theoretical) overhead
+            # of the the ``BroadcastReceiver`` before calling
+            # this method for the first time.
+
+            # XXX: why does this work without a recursion issue?!
```
goodboy, line 299: This was broken actually, but is now fixed by adding support for passing into `BroadcastReceiver` a `recv_afunc` kwarg that is by default the underlying's `.receive()` if not provided.

This makes it easy to wrap receive channel apis for use in task broadcasting without expecting any initial consumer-task of the underlying to also use a broadcast receiver instance when calling `.receive()`, it however does mutate the machinery to put the broadcaster *behind the scenes* as it were.

goodboy:
Oh right, to use `typing.Protocol` we need 3.8+. I wonder is this incentive enough to drop 3.7 for next alpha?

tests/test_local_task_broadcast.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # make sure it all works within the runtime
+        async with tractor.open_root_actor():
+
+            num_laggers = 4
+            laggers: dict[str, int] = {}
+            retries = 3
+            size = 100
+            tx, rx = trio.open_memory_channel(size)
+            brx = broadcast_receiver(rx, size)
```
goodboy, line 121: I'm almost tempted to enforce this as an async context manager, something I kinda wish `open_memory_channel()` did as well.

tests/test_local_task_broadcast.py
```diff
@@ -0,0 +1,222 @@
...
+            async def sub_and_print(
+                delay: float,
+            ) -> None:
+
+                task = current_task()
+                start = time.time()
+
+                async with brx.subscribe() as lbrx:
+                    while True:
+                        #     await tractor.breakpoint()
```
goodboy, line 132: remove.

tractor/_broadcast.py
```diff
@@ -0,0 +1,275 @@
...
+    # on a newly produced value from the sender.
+    recv_ready: Optional[tuple[str, trio.Event]] = None
+
+
+class BroadcastReceiver(ReceiveChannel):
+    '''A memory receive channel broadcaster which is non-lossy for the
+    fastest consumer.
+
+    Additional consumer tasks can receive all produced values by registering
+    with ``.subscribe()`` and receiving from thew new instance it delivers.
```
goodboy, line 87: typo `thew`

tractor/_broadcast.py
```diff
@@ -0,0 +1,275 @@
...
+        '''Subscribe for values from this broadcast receiver.
+
+        Returns a new ``BroadCastReceiver`` which is registered for and
+        pulls data from a clone of the original ``trio.abc.ReceiveChannel``
+        provided at creation.
+
+        '''
+        # use a uuid4 for a tee-instance token
+        key = str(uuid4())
+        state = self._state
```
goodboy, line 221: We should probably also error out if `self._closed` and it's a new `.subscription()` entry?

tractor/_broadcast.py
```diff
@@ -0,0 +1,275 @@
...
+        finally:
+            await br.aclose()
+
+    # TODO:
+    # - should there be some ._closed flag that causes
+    #   consumers to die **before** they read all queued values?
+    # - if subs only open and close clones then the underlying
+    #   will never be killed until the last instance closes?
+    #   This is correct right?
+    async def aclose(
```
goodboy, line 242: This doesn't need to be async now which means neither does `.subscribe()`.
The only nice thing about it being async is the [parentheses syntax](https://docs.python.org/3.10/whatsnew/3.10.html#parenthesized-context-managers) coming in 3.10 which is likely where the entry could be inlined with `Context.open_stream()`.

tractor/_broadcast.py
```diff
@@ -0,0 +1,275 @@
...
+
+        if self._closed:
+            return
+
+        # XXX: leaving it like this consumers can still get values
+        # up to the last received that still reside in the queue.
+        # Is this what we want?
+        self._state.subs.pop(self.key)
+        # if not self._state.subs:
+        #     await self._rx.aclose()
```
goodboy, line 254: This shouldn't get touched right?

It's up to the caller to determine when the underlying should be closed?

tractor/_streaming.py
```diff
@@ -253,6 +259,48 @@ async def aclose(self):
...
+                self._rx_chan._state.max_buffer_size,  # type: ignore
+               receive_afunc=self.receive,
+            )
+
+            # NOTE: we override the original stream instance's receive
+            # method to now delegate to the broadcaster's ``.receive()``
+            # such that new subscribers will be copied received values
+            # and this stream doesn't have to expect it's original
+            # consumer(s) to get a new broadcast rx handle.
+            self.receive = bcast.receive  # type: ignore
```
goodboy, line 297: was just thinkin, could we revert this on exit?

does it require waiting for the last sub to also be closed?
- if `broadcast_receiver()` was a ctx mngr this would easier
- is this too *monkey-patchy*?

tractor/_streaming.py
```diff
@@ -269,17 +317,6 @@ async def send(
         '''
         await self._ctx.chan.send({'yield': data, 'cid': self._ctx.cid})
 
-    # TODO: but make it broadcasting to consumers
-    def clone(self):
```
goodboy, line 273: Good riddance.

tractor/_broadcast.py
goodboy: Yeah sticking with underlying lifetime mgmt being up to use code.

goodboy:
This latest test should cause failures.

Commit coming to fix it once that's verified.

goodboy:
Ok or it'll just work without the fix 🙄

tractor/_broadcast.py
```diff
@@ -0,0 +1,308 @@
...
+                raise trio.ClosedResourceError
+
+            event = trio.Event()
+            state.recv_ready = key, event
+
+            # if we're cancelled here it should be
+            # fine to bail without affecting any other consumers
+            # right?
+            try:
+                value = await self._recv()
```
goodboy, line 177: My last edge case concern in the one where this call gets cancelled and we leave the `state.recv_ready` event consumers potentially stuck. What should happen then?

If we wake them up then we need one of them to re-enter and make this call again.
What happens to the others?
They have to be placed back in the event `.wait()`

In theory the below commented recursive call would solve this right?
Is it necessary tho?

tractor/_broadcast.py
```diff
@@ -0,0 +1,308 @@
...
+
+            # if seq > -1:
+            #     # stuff from above..
+            # elif seq == -1:
+            #     # XXX: In the case where the first task to allocate the
+            #     # ``.recv_ready`` event is cancelled we will be woken with
+            #     # a non-incremented sequence number and thus will read the
+            #     # oldest value if we use that. Instead we need to detect if
+            #     # we have not been incremented and then receive again.
+            #     return await self.receive()
```
goodboy, line 246: In theory this might solve the edge case mentioned above.

I'm not sure why the one test isn't triggering this case; might need a more specific test..

tractor/_broadcast.py
```diff
@@ -0,0 +1,308 @@
...
+        # This task is all caught up and ready to receive the latest
+        # value, so queue sched it on the internal event.
+        else:
+            seq = state.subs[key]
+            assert seq == -1  # sanity
+            _, ev = state.recv_ready
+            await ev.wait()
+
+            seq = state.subs[key]
+            assert seq > -1, f'Invalid sequence {seq}!?'
```
goodboy, line 227: True since currently we'll never awake sibling consumers if the first-task-to-.receive() is cancelled.

goodboy:
Gonna try to land #230 first so the tests here need some tweaks.

goodboy:
Hah! so i was right. the cancel on first receiver case was borked.
Last commit there fixes it 😎

PR #230, opened 2021-09-01
Drop stream shielding; it was from a legacy api design

goodboy:
Before we had `async with Portal.open_stream_from()` which has explicit enter/exit semantics, we had a implicit *portal-returns-stream-reference* thing which required that if a `stream.receive()` was cancelled, we relayed that via a stream closure to the other side.

Since we have the new API now we don't need this shielding (aka we won't `.aclose()` the stream on cancellation) any more.

This PR removes all that and the CI should prove that all things are well. 

Once CI is clean I'll just remove the first patch draft which comments all the unneeded code.

### DIFF
--- a/tractor/_streaming.py
+++ b/tractor/_streaming.py
@@ -46,11 +46,9 @@ class ReceiveMsgStream(trio.abc.ReceiveChannel):
         self,
         ctx: 'Context',  # typing: ignore # noqa
         rx_chan: trio.abc.ReceiveChannel,
-        shield: bool = False,
     ) -> None:
         self._ctx = ctx
         self._rx_chan = rx_chan
-        self._shielded = shield
 
         # flag to denote end of stream
         self._eoc: bool = False
@@ -103,7 +101,10 @@ class ReceiveMsgStream(trio.abc.ReceiveChannel):
         except (
             trio.ClosedResourceError,  # by self._rx_chan
             trio.EndOfChannel,  # by self._rx_chan or `stop` msg from far end
-            trio.Cancelled,  # by local cancellation
+
+            # Wait why would we do an implicit close on cancel? THAT'S
+            # NOT HOW MEM CHANS WORK!!?!?!?!?
+            # trio.Cancelled,  # by local cancellation
         ):
             # XXX: we close the stream on any of these error conditions:
 
@@ -135,23 +136,6 @@ class ReceiveMsgStream(trio.abc.ReceiveChannel):
 
             raise  # propagate
 
-    @contextmanager
-    def shield(
-        self
-    ) -> Iterator['ReceiveMsgStream']:  # noqa
-        """Shield this stream's underlying channel such that a local consumer task
-        can be cancelled (and possibly restarted) using ``trio.Cancelled``.
-
-        Note that here, "shielding" here guards against relaying
-        a ``'stop'`` message to the far end of the stream thus keeping
-        the stream machinery active and ready for further use, it does
-        not have anything to do with an internal ``trio.CancelScope``.
-
-        """
-        self._shielded = True
-        yield self
-        self._shielded = False
-
     async def aclose(self):
         """Cancel associated remote actor task and local memory channel
         on close.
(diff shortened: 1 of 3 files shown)


tractor/_streaming.py
```diff
@@ -169,18 +153,6 @@ async def aclose(self):
...
-        #     # if we've been cloned don't kill the stream
-        #     log.debug(
-        #       "there are still consumers running keeping stream alive")
-        #     return
-
-        if self._shielded:
-            log.warning(f"{self} is shielded, portal channel being kept alive")
-            return
-
         # XXX: This must be set **AFTER** the shielded test above!
```
richardsheridan, line 156: This XXX note is obsolete now?  also, what happens when `await self._ctx.send_stop()` below is cancelled? I can't tell locally if that is handled internally, or for that matter if cancelling it is not a problem.

goodboy: Ahh good catch and good question.

Technically if we're cancelled there's not a lot that can be done anyway since you run into the 2-general's problem in terms of getting back and ack to the `stop` msg.  So even if we were to wait, how long would we wait, and further if we're cancelled the transport layer may still stay active but the far end task may have already terminated (depending on which side of the stream was the context opener).

One thing we can think about is a stricter ack to the surrounding context cancellation maybe?

goodboy: Yeah so i think the short answer is, if we're cancelled on `.send_stop()` then it's as if it wasn't sent and the surrounding context is responsible for relaying that cancellation back to the other side.

richardsheridan: a similar call is cancelled there, so it seems like the far side is abandoned IIUC. to maintain your one-shot invariant you'll need a shielded scope with a deadline, and some way to configure that deadline, and then some way to mark the other side as lost (and maybe panic and blow up the whole portal?) when the deadline passes.

goodboy: Ahh yes sorry so in the bidir case this block isn't invoked on the *caller* (opener of context) side since `self._ctx._portal` will be `None`. So if the far end *callee* task is cancelled while closing,  i'm not sure it improves anything to *try-to-ensure* the `stop` is relayed given that we'll get the context cancel bubble across anyway on the caller side. Maybe this is not clear enough or should change though?

goodboy: > so it seems like the far side is abandoned IIUC.

The caller is unaware of the stream termination until the cancel bubbles up on the callee side yes.
When that eventually happends we trigger the [invoke machinery](https://github.com/goodboy/tractor/blob/master/tractor/_actor.py#L152) and the context should be cancelled by another message.

So it can't really be *abandoned* per say bc the context is in charge of that, however, yes the state of the stream's closure is unknown.

goodboy: So i guess the good question you've made is *should there be attempted stream state consistency* on cancellation during `.aclose()` yes?

richardsheridan: right and it seems to me that's fine as long as it's consistent with the rest of the system design.

goodboy: yeah, this was the whole point of `Portal.open_context()` being necessary **before** you can open a stream - exactly to handle the cancellation/failure case (in one place) without having to try and jam it into the stream state tracking.

richardsheridan: as in, does any other part of the system need to know if someone wanted to close the stream but couldn't do it in time? probably not but I wouldn't know.

goodboy: Yeah i don't *think* so.

As per discussion in chat, if you have a stream state desync there's not a ton can be done without a supervisory strategy to handle the cancellation (could be parent(caller) or child(callee) side) and either way, the stream shouldn't be closed implicitly by cancellation (what this patch changes) so it's up to whoever handles that `trio.Cancelled` to determine what to do about a cancellation - in the most default case the stream is closed more thoroughly in the exit step and later in the context closure.

For our current support of only a *one shot streaming phase* in the dialog i think this is sufficient since we shouldn't be attempting to re-open the same stream without a new context.

PR #234, opened 2021-09-02
Root tty hangs

goodboy:
Debugger improvements to avoid even further tty clobbering edge cases and potential root actor hangs when the debug lock is acquired but the child completes or intends to release it but the root never is notified.

### DIFF
--- a/tractor/_entry.py
+++ b/tractor/_entry.py
@@ -62,7 +62,7 @@ def _trio_main(
     """
     # Disable sigint handling in children;
     # we don't need it thanks to our cancellation machinery.
-    signal.signal(signal.SIGINT, signal.SIG_IGN)
+    # signal.signal(signal.SIGINT, signal.SIG_IGN)
 
     log.info(f"Started new trio process for {actor.uid}")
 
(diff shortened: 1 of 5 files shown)


goodboy:
Need to squash / drop a few of these commits.

tractor/_trionics.py
```diff
@@ -298,8 +298,8 @@ async def _open_and_supervise_one_cancels_all_nursery(
                                 f'child {_debug._global_actor_in_debug}\n'
                                 'Waiting on tty lock to release..')
 
-                            with trio.CancelScope(shield=True):
-                                await debug_complete.wait()
+                            # with trio.CancelScope(shield=True):
```
goodboy, line 301: We may end up wanting to keep this but right now it's causing too many hangs.

tractor/_entry.py
```diff
@@ -62,7 +62,7 @@ def _trio_main(
     """
     # Disable sigint handling in children;
     # we don't need it thanks to our cancellation machinery.
-    signal.signal(signal.SIGINT, signal.SIG_IGN)
+    # signal.signal(signal.SIGINT, signal.SIG_IGN)
```
goodboy, line 65: Might as well keep for reference.

PR #239, opened 2021-10-04
Fix kbi in ctx block

goodboy:
Fix for an issue discovered in the development of [`wrath`](https://github.com/adder46/wrath) - thanks to @adder46!

It turns out we need to catch KBI's in `Portal.open_context()` blocks in order to get proper cancellation *if* the block is run inside the top level task (maybe always).  The fix for this is simple, we just need to either explicitly handle the kbi in the block's `__aenter__()` implementation or just handle `BaseException` (which is the likely solution).

Tests are up first to show CI failures, patch will come shortly after.

### DIFF
--- a/tractor/_portal.py
+++ b/tractor/_portal.py
@@ -177,7 +177,6 @@ class Portal:
                 f"Cancelling all streams with {self.channel.uid}")
             for stream in self._streams.copy():
                 try:
-                    # with trio.CancelScope(shield=True):
                     await stream.aclose()
                 except trio.ClosedResourceError:
                     # don't error the stream having already been closed
@@ -294,7 +293,6 @@ class Portal:
     async def open_stream_from(
         self,
         async_gen_func: Callable,  # typing: ignore
-        shield: bool = False,
         **kwargs,
 
     ) -> AsyncGenerator[ReceiveMsgStream, None]:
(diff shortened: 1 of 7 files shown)


goodboy:
Nice, [CI just caught](https://github.com/goodboy/tractor/pull/239/checks?check_run_id=3793544086#step:5:271) the `.open_stream()` race condition where the parent opens but the child does not and received a `'stop'` msg which it doesn't recognize. We need an issue and solution for how to handle that case.

tractor/_portal.py
```diff
@@ -427,9 +434,14 @@ async def open_context(
                 log.debug(f'Context {ctx} cancelled gracefully')
 
         except (
-            trio.Cancelled,
-            trio.MultiError,
-            Exception,
+            BaseException,
```
goodboy, line 437: This is the main change that gets us the fix.
The alternative would be to include the kbi in the old error set below but it's likely that any other base exception should trigger this same cancellation behavior afaik.

tractor/_streaming.py
```diff
@@ -396,6 +398,10 @@ async def cancel(self) -> None:
                     log.warning(
                         "May have failed to cancel remote task "
                         f"{cid} for {self._portal.channel.uid}")
+                else:
+                    log.warning(
```
goodboy, line 402: This is just new logging, yes noisy but handy until this cancel stuff is better refined imo.

We can decrease the level if complaints pile in.

Issue #242, opened 2021-10-05
Spawning a cluster of actors


xdb64:
I'm using `tractor` with my [TCP SYN port scanner](https://github.com/adder46/wrath) and I'm amazed by its smooth integration with trio, as well as the performance I've managed to squeeze out of my computer using these great libs.

During the development, I've come across the need to spawn a cluster of actors, open contexts on all of them, and then open streams on those. Currently, I have to hard-code everything (one spawn-context-stream triplet per actor) which results in a big `async with` block. I wish there was a handy utility procedure that would spawn `n` actors, open contexts on all of them, and finally, open streams - thus allowing for a number of streams proportional to the number of CPU cores to be created, for example. I'm sure what I'm asking is too specific, but I am also sure that these peculiarities could be abstracted away in a couple of generalized utility routines for doing similar tasks.

I know some of this is on its way, but in any case, please let me know once it's ready; I'd be happy to assist with battle-testing.

goodboy:
>  I'm amazed by its smooth integration with trio, as well as the performance I've managed to squeeze out of my computer using these great libs.

😎 

@adder46 thanks for the issue friend.

I'm hoping we can get you what you need in #241 and ideally get some decent teardown latency going as well.

As mentioned in chat, the `Context.open_stream()` only really benefits from concurrency on `__aexit__()` (since it'll send a `'stop'` msg on close) so we'll need to target maybe even avoiding that bit if we want some kind of *hard shutdown* of streams.

goodboy:
I *think* we added this in #241 yah?

goodboy:
Closed via #241.

PR #246, opened 2021-10-17
Patch async enter all (addresses #242)

overclockworked64:
Ready for review.

There was an issue with getting current actor's UID because at the time we try to get it, no actor is currently spawned (at least that's what the error said), so I left a random UID as a temporary workaround - this should probably be addressed. Another thing is that I thought that it's more appropriate for `async_enter_all` to take a sequence of managers. A couple of mypy errors were fixed as well. Finally, a test that demonstrates new capabilities was added.

Minor nitpick: we should enforce some kind of consistent neutral code formatting style across the entire codebase.

### DIFF
--- a/tests/test_clustering.py
+++ b/tests/test_clustering.py
@@ -23,19 +23,15 @@ async def worker(ctx: tractor.Context) -> None:
 
 @tractor_test
 async def test_streaming_to_actor_cluster() -> None:
-    teardown_trigger = trio.Event()
     async with (
         open_actor_cluster(modules=[__name__]) as portals,
         async_enter_all(
             mngrs=[p.open_context(worker) for p in portals.values()],
-            teardown_trigger=teardown_trigger,
         ) as contexts,
         async_enter_all(
             mngrs=[ctx[0].open_stream() for ctx in contexts],
-            teardown_trigger=teardown_trigger,
         ) as streams,
     ):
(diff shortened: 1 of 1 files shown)


goodboy:
@overclockworked64 thanks for these fixes!

Glad to see you got something working for `wrath` 😎 

>  I left a random UID as a temporary workaround - this should probably be addressed.

Where does this issue stem from? Are we trying to do something before the runtime comes up a la `open_root_actor()`?

> more appropriate for async_enter_all to take a sequence of managers

Yeah, fine by me. Sequence in, sequence out would be more ergonomic I'd think.

> Finally, a test that demonstrates new capabilities was added.

🏄🏼 

> Minor nitpick: we should enforce some kind of consistent neutral code formatting style across the entire codebase.
Yeah, I've changed my position on style a few times since this code base was started - which is probably what you're seeing 😂.

Thoughts from you and lurkerz are (mostly) welcome.

overclockworked64:
> > I left a random UID as a temporary workaround - this should probably be addressed.
> 
> Where does this issue stem from? Are we trying to do something before the runtime comes up a la `open_root_actor()`?

As is, this code gives a `RuntimeError`, I'm assuming probably because we're trying to do something with the current actor before actually opening a nursery and spawning some actors.

PR #257, opened 2021-10-27
Add `maybe_open_context()` an actor wide task-resource cache

goodboy:
Like it sounds: a context manager that only opens a new actor-global (aka mem local) resource if one isn't already cached.

This is particularly handy for resources that can be shared by multiple tasks in and actor where the user desires that the lifetime of the resource is maintained until the final *user task* completes.

Still needs tests obviously.

UPDATE:
This ended up turning into a foray into fixing `MsgStream` closure/termination semantics after trying to get the original resource caching test working.

Todo tests:
- [x] cached inter-actor stream demonstration
- [x] simpler audit of just ensuring that multiple local `trio` tasks get the same cached value
- [x] error/cancel tests?

### DIFF
--- a/newsfragments/257.feature.rst
+++ b/newsfragments/257.feature.rst
@@ -1,0 +1,9 @@
+Add ``trionics.maybe_open_context()`` an actor-scoped async multi-task
+context manager resource caching API.
+
+Adds an SC-safe cacheing async context manager api that only enters on
+the *first* task entry and only exits on the *last* task exit while in
+between delivering the same cached value per input key. Keys can be
+either an explicit ``key`` named arg provided by the user or a
+hashable ``kwargs`` dict (will be converted to a ``list[tuple]``) which
+is passed to the underlying manager function as input.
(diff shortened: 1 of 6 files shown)


goodboy:
@overclockworked64 there I think the test should explain a little better how this works.

I also fixed up the naming concern to `_Cache`.

goodboy:
gahhhhh windowwwwws

goodboy:
Lol this `mypy` thing..

```
mypy.ini:2: error: Error importing plugin "trio_typing.plugin": cannot import name 'TypeVarDef' from 'mypy.types' (/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/mypy/types.cpython-39-x86_64-linux-gnu.so)
```

goodboy:
If the streaming improvements here seem too *lumped in* we can also break them out into a separate PR.

It might be better to more formally document those fixes and the new test?
I dunno wat y'all think?

PR #258, opened 2021-11-01
Add a @pub kwarg to allow specifying a "startup response message"

goodboy:
This is a fix for an (old and) undocumented feature that enabled async generator based "pub-sub" streaming such that during stream connection startup an initial "message" is sent to each new subscriber much like how the new `tractor.Context.started()` works for bidirectional streaming setup.

This patch in particular was pulled out of historical work on the `infect_asyncio` branch.

I'm not even sure this undocumented [internal "pub sub" api](https://github.com/goodboy/tractor/blob/master/tractor/msg.py#L120) should even be kept since it is basically just a *multiplexed* (read non-task oriented) version of what you can already do easily with our task-broadcast receiver apis and a single producer actor.

It's probably worth making a so called *dynamic pub-sub* example using the newer APIs for comparision, maybe do some brief perf benchmarks and then decide if we should just drop this interface?

### DIFF
--- a/tractor/msg.py
+++ b/tractor/msg.py
@@ -121,6 +121,7 @@ def pub(
     wrapped: typing.Callable = None,
     *,
     tasks: Set[str] = set(),
+    send_on_connect: Any = None,
 ):
     """Publisher async generator decorator.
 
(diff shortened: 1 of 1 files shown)


goodboy:
I'm thinking we should do a rewrite of this using the new bi-dir streaming + `MsgStream.subscribe()` apis and completely drop this internal subsystem.

PR #259, opened 2021-11-01
Alpha3

goodboy:
3rd alpha release.
Also tossed in 3.10 to CI let's see if that works 😂 

Notes for *trio montly*, ping @Fuyukai:

`tractor` saw its 3rd alpha release with some interesting improvements in the core cross-actor cancellation mechanics as well as the start of a public expose of what were our internal `trio` helper API goodies and we officially now have optional  `msgspec` support!

The quick feature summary:
- We switched to making windows use the `trio`/`subprocess` spawner by default which means windows users now get multi-core debugger support out of the box.
- Support for optional [`msgspec`](https://github.com/jcrist/msgspec) lib usage for those who want a faster and typed codec alternative to [`msgpack-python`](https://github.com/msgpack/msgpack-python).
- We exposed a new sub-package `tractor.trionics` which is where we'll be stashing all our "higher level" `trio` related helper APIs going forward. For example checkout our new [`gather_contexts`](https://github.com/goodboy/tractor/blob/master/tractor/trionics/_mngrs.py#L39) concurrent context manager enter-er (kinda like a concurrent version of `asyncio.gather()` but for context manager instances).
- Dropped Py 3.8 support as we decided to roll with a new model of supporting only the latest 2 major releases which means we also now support 3.10.
- Change the core message loop to handle task and actor-runtime cancel
requests **immediately** instead of scheduling them as is done for rpc-task
requests. Previously cancel requests were actually *scheduled* as rpc requests but this has changed to instead invoke a synchronous style cancellation which resulted in a massive reduction in teardown race conditions particularly in our anti-zombie-process machinery and it's interaction with the debugger system.


guilledk (approved):

overclockworked64 (approved):

goodboy:
There super naively dropped 3.10 on windows for now which I'm thinking is nbd for this release (because it likely works just fine).

goodboy:
Ugh but mistakenly dropped msgspec testing for windows..
We know it works (due to previous runs) so not going to stress about fixing CI before this release.

PR #260, opened 2021-11-02
Clusters and hot tips

goodboy:
Adds an example and mirrors it in the readme.

I tweaked the auto-generated names for actors created with `open_actor_cluster()` to be simpler (since they had this weird `.uid` string thing going on) to be simply `<parent_name>.<name>`.

Oh right, and started a little internal *dev tips* doc just for reminding myself of stuff.
It needs lots more content don't worry.

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -313,8 +313,68 @@ real time::
 This uses no extra threads, fancy semaphores or futures; all we need
 is ``tractor``'s IPC!
 
+To be extra terse the ``tractor`` devs have started hacking some "higher
+level" APIs for managing actor trees/clusters. These interfaces should
+generally be condsidered provisional for now but we encourage you to try
+them and provide feedback. Here's a new API that let's you quickly
+spawn a flat cluster:
+
+.. code:: python
+
+    from functools import partial
+
+    import trio
+    import tractor
+
+
+    async def sleepy_jane():
+        uid = tractor.current_actor().uid
+        print(f'Yo i am actor {uid}')
+        await trio.sleep_forever()
+
+
+    async def main():
+        '''
+        Spawn a flat actor cluster, with one process per
+        detected core.
+
+        '''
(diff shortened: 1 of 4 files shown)


tractor/_clustering.py
```diff
@@ -38,7 +38,7 @@ async def open_actor_cluster(
             uid = tractor.current_actor().uid
 
             async def _start(name: str) -> None:
-                name = f'{name}.{uid}'
```
goodboy, line 41: This looked super horrible/confusing if you actually printed it 😂

overclockworked64 (approved):

docs/README.rst
```diff
@@ -313,9 +313,69 @@ real time::
...
+            tractor.open_actor_cluster(
+                modules=[__name__]
+            ) as portal_map,
+
+            trio.open_nursery() as n,
+        ):
+
+            for (name, portal) in portal_map.items():
+                n.start_soon(
+                    partial(
```
goodboy, line 357: just realizing i guess we don't actually need the `partial` right?

examples/quick_cluster.py
```diff
@@ -0,0 +1,50 @@
...
+        tractor.open_actor_cluster(
+            modules=[__name__]
+        ) as portal_map,
+
+        trio.open_nursery() as n,
+    ):
+
+        for (name, portal) in portal_map.items():
+            n.start_soon(
+                partial(
```
goodboy, line 34: Yah, I had the `partial()` originally because I thought we needed to pass a `name='blah'` kwarg but we're actually auto-generating them unless a sequence is passed.

I'm wondering if we should tweak this api to allow passing in *only* the `names: list[str]` and if so then the helper makes only as many actors as there are names? Also, in the case where `names` **is** passed we probably shouldn't be doing this `<parentname>.<childname>` generation right?

PR #261, opened 2021-11-05
`Context` oriented error relay and `MsgStream` overruns 

### LINKED ISSUE #265: Feeder channel lookup race condition during context setup
this test sometimes catches it: `tests/test_task_broadcasting.py .........F..`, also seen in this [actions run](https://github.com/goodboy/tractor/runs/4346868417?check_suite_focus=true).

Main issue traceback:

```python
Nov 28 19:07:04 (root, 1927, trio._highlevel_serve_listeners._run_handler)) [ERROR] tractor _actor.py:794 Actor errored:
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/tractor/_actor.py", line 666, in _process_messages
    ns, funcname, kwargs, actorid, cid = msg['cmd']
KeyError: 'cmd'

During handling of the above exception, another exception occurred:

  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/tractor/_actor.py", line 676, in _process_messages
    raise exc
tractor._exceptions.RemoteActorError: ('sequence_echoer', '92bdaa8f-9824-434f-8536-45a96a63fec7')
  File "/home/runner/work/tractor/tractor/tractor/_actor.py", line 658, in _process_messages
    await self._push_result(chan, cid, msg)
  File "/home/runner/work/tractor/tractor/tractor/_actor.py", line 547, in _push_result
    send_chan, recv_chan = self._cids2qs[(chan.uid, cid)]
KeyError: (('root', '44b81b41-b860-49d5-a1d4-a47937d77763'), '4184a753-574e-4044-8d49-dfb4f895e25b')
```
---
Description:
pretty sure this race is where there is a new message that arrives from the "caller" task (in this case in the root actor) but the target "callee" hasn't yet called `tractor.Context.open_stream()` which [internally enters the feeder mem chan into the rpc lookup map](https://github.com/goodboy/tractor/blob/master/tractor/_streaming.py#L463).

The question is how to handle this?
- we could queue the message expecting the callee to eventually look it up?
  - seems cludgy, and what would be the reason to **not** have a mem chan feeder other then for *one way streaming*?
  -  but, do we care about pedantically avoiding an extra mem chan in that case?
- we just enter the feed chan at [`_invoke()` time](https://github.com/goodboy/tractor/blob/master/tractor/_actor.py#L155) (seems to me to be the most correct right?)

---
TODO:
- a test that demonstrates this reliably which should be something like a `ctx.started()` followed by a delay before a `ctx.open_stream():` and the caller should send during that delay period.
- probably coordinate this with https://github.com/goodboy/tractor/pull/261 and accompanying tests

goodboy:
Stricter semantics around `tractor.Context.open_stream():` and friends:

- Previously we were ignoring a race where the callee an opened task
context could enter `Context.open_stream()` before calling `.started()`.
- Disallow this as well as calling `.started()` more then once.
- Support a new `StreamOverrun` error that is raised in 2 main cases:
  - `Context.open_stream()` is entered on only one side and that side sends more messages then the `Actor.msg_buffer_size` in which cases the error is send back to the sending side with a special warning if the receive side never entered its own `.open_stream()`
  - `Context.open_stream(backpressure=False)` is explicitly set by the user in which case the error may show up at any `trio` checkpoint (via embedding in a `tractor.RemoteActorError`) on the send side if an overrun takes place. In this case user code expected to handle the error if they don't want a send side task crash - hence why this is **not** the default setting.

Tests todo:
- [x] context started > once
- [x] `.open_stream()` entered before `.started()` called
- [x] test(s) that verify #265 is fixed
- one side of stream not opened but is overrun by other side:
  - [x] caller overruns callee
  - [x] callee overruns caller
  - [x] overrun managed via backpressure  

Races that inspired this discovered as part of https://github.com/pikers/piker/pull/241

Fixes #265


goodboy:
Obvs this needs some unit tests.

goodboy:
Added a test that we need a fix for re: #265.

goodboy:
@guilledk @overclockworked64 I think this one is ready to land.

There might be one last little test I forgot - gotta check.

Oh and we need nooz.. might be a long one 😂

guilledk (approved):

PR #263, opened 2021-11-08
Early deth fixes

goodboy:
A bug fix for debug lock acquirer helper and a fix for the early actor termination case where the TCP server never fully comes up before the runtime is killed.

### DIFF
--- a/tractor/_actor.py
+++ b/tractor/_actor.py
@@ -248,7 +248,7 @@ async def _invoke(
                 # If we're cancelled before the task returns then the
                 # cancel scope will not have been inserted yet
                 log.warning(
-                    f"Task {func} likely errored or cancelled before it started")
+                    f"Task {func} likely errored or cancelled before start")
         finally:
             if not actor._rpc_tasks:
                 log.runtime("All RPC tasks have completed")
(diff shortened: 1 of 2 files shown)


goodboy:
Y'all think we need news fragments for this?

tractor/_debug.py
```diff
@@ -557,8 +557,13 @@ async def acquire_debug_lock(
     '''
     Grab root's debug lock on entry, release on exit.
 
+    This helper is for actor's who don't actually need
+    to acquired the debugger but want to wait until the
+    lock is free in the tree root.
+
     '''
     if not debug_mode():
+        yield None
```
goodboy, line 566: 2nd bug fix. This was a very minor missing `yield`.

tractor/_actor.py
```diff
@@ -1123,7 +1135,11 @@ async def cancel(self) -> bool:
 
             # stop channel server
             self.cancel_server()
-            await self._server_down.wait()
+            if self._server_down is not None:
+                await self._server_down.wait()
```
goodboy, line 1139: Handles the case where the TCP server task never comes online before a cancellation request is serviced from the parent.

goodboy (commented):
Pretty Sure this is all minor enough to just merge silently.

PR #266, opened 2021-11-29
Faster graceful daemon cancels

goodboy:
This was *kind of* a bug where if the soft wait (`await .proc.wait()`) in the reaper task was cancelled we could fail over to the *hard reap sequence* which has a timeout of 3s (currently) before system signalling the subproc.

This adds a test to catch that *slower then necessary* reaping delay and adds the fix where if the graceful daemon actor proc wait is cancelled we relay that cancellation to the actor via `Portal.cancel_actor()` which should result in a much faster graceful teardown.

I'm pretty sure I already worked through some of this in the (now history mangled)  [zombie lord](https://github.com/goodboy/tractor/commits/zombie_lord_infinite) branch..

### DIFF
--- a/newsfragments/266.bug.rst
+++ b/newsfragments/266.bug.rst
@@ -1,0 +1,12 @@
+Fix graceful cancellation of daemon actors
+
+Previously, his was a bug where if the soft wait on a sub-process (the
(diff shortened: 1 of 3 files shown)


goodboy:
ugh, frickin windows.

PR #271, opened 2021-12-07
Debug flag per actor

goodboy:
Draft try at allowing per-actor debug mode control(s). 

Also some fixes to the test suite to actually use correct log level settings as well as some logic to use `PDB` level logging by default whenever `debug_mode=True` is set.

This this probably could use some tests?


goodboy:
@guilledk besides nooz, if CI is clean, i think this is ready to land as well.

Issue #103, opened 2020-02-05
Change our license?


goodboy:
I'm not sold on GPLv3 (it was a hip shot to get things going with what little I knew).
I want to open discussion on this.

One thing I do want to enforce is that any corp trying to use this project for profit pays back into it; if you're using it for free and sharing what you add then 100% it's acceptable and we should propagate that use.

With that very naive principle in mind here are some options that others have pointed me to:
- AGPLv3 used by the [Genode project](https://genode.org/about/licenses)
- [license zero](https://licensezero.com/) which has [an interesting public-private approach](https://guide.licensezero.com/) more along the lines of `Qt` or `MongDB`

I'm of course open to other suggestions (even for moving to looser more well known options) with good arguments for why.

parity3:
I don't have any opinions on this but thought I'd add a couple recent references to software licensing I've come across:

[sentry blog post](https://blog.sentry.io/2019/11/06/relicensing-sentry)
[changelog podcast interview episode talking about sentry's licensing strategy](https://changelog.com/podcast/371)

Obviously, this does not pertain directly to tractor, but if you're trying to find out where this segment of open source is going, particularly if you ever want to run a SaaS on any of your FOSS in the future (which probably is not the case), I think it's a good data point.

goodboy:
Sweet. Thanks for the input @parity3 :+1:

ryanhiebert:
It doesn't meet your stated goal "to enforce is that any corp trying to use this project for profit pays back into it", but since you allowed for looser ones with arguments of why, I prefer MIT. Its minimal, simple, and free-free.

The reason that I prefer MIT is because copyright law, it seems to me, is fundamentally broken. It is well-known that copy-left licenses like GPL simply would not work if software copyright were eliminated. Combine that with the minimal, simple, and well-known nature of the MIT license, and that establishes my preference, even though my thinking on copyright isn't quite so simple as "it needs to be abolished."

I do think that the recent changes to Sentry licensing, as well as that license zero link (which I'd never seen before) make for interesting thinking, and I'm interested to see what novel and more ethical business models may come as a result.

I am concerned that requiring for-pay organizations to pay for a license to tractor would have the effect of limiting the ease by which they may adopt tractor, and that it may end up backfiring with regard to overall adoption of tractor. In my company, for example, we'd have to go through some management overhead to start working with tractor, while a free license would not require any approval from management to adopt.

[...]

PR #274, opened 2021-12-13
Re-license code base for distribution under AGPLv4

goodboy:
This commit obviously denotes a re-license of all applicable parts of
the code base (not including the minor patch setl not yet verified by
contributors). From here henceforth all changes will be AGPLv4 licensed
and distributed. This is purely an effort to maintain the same copy-left
policy whilst closing the SaaS loophole the GPLv4 allows for. It is
merely for this loophole, to avoid code hiding by any potential firms
who are attempting to use the project to make a profit without either
compensating the authors or re-distributing their changes.

I thought quite a bit about this change and can't see a reason not to
close the SaaS loophole in our current license. We still are (hard)
copy-left and I plan to keep the code base this way for a couple
reasons:

- The code base produces income/profit through parent projects and is
  demonstrably of high value.
- I believe firms should not get free lunch for the sake of
  "contributions from their employees" or "usage as a service" which
  I have found to be a dubious argument at best.
- If a firm who intends to profit from the code base wants to use it
  they can propose a commercial license to purchase with the proceeds
  going to the project's authors under some form of well defined
  contract.
- Many successful projects like Qt use this model; I see no reason it
  can't work in this case until such a time as the authors feel it
  should be loosened.

There has been detailed discussion in #103 on licensing alternatives.
The main point of this AGPL change is to protect the code base for the
time being from exploitation and hijack as we move into the next phase
of development which will include extending into the multi-host
distributed software space.

-----
ping @guilledk @chrizzFTD @overclockworked64.

As all of you are contributors with [varying levels of patches](https://github.com/goodboy/tractor/graphs/contributors) to the codebase this is a formal reach out to ask for approval to re-license and distribute henceforth each of your changes as the code base is continued to be developed into the the future.

You are of course not expected in any way to re-license your changes but know that changes to your changes will be further more redistributed under the new license.

Cheers!

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,23 +1,21 @@
-                    GNU GENERAL PUBLIC LICENSE
-                       Version 3, 29 June 2007
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
 
- Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 
                             Preamble
 
-  The GNU General Public License is a free, copyleft license for
-software and other kinds of works.
+  The GNU Affero General Public License is a free, copyleft license for
+software and other kinds of works, specifically designed to ensure
+cooperation with the community in the case of network server software.
 
   The licenses for most software and other practical works are designed
 to take away your freedom to share and change the works.  By contrast,
-the GNU General Public License is intended to guarantee your freedom to
+our General Public Licenses are intended to guarantee your freedom to
 share and change all versions of a program--to make sure it remains free
-software for all its users.  We, the Free Software Foundation, use the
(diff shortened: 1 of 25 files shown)


guilledk:
The following is my opinion only:

In an ideal world no intellectual property law would exist and real free market competition would ensue.
But in reality those systems are well put in place and there are those who routinely abuse them. 

With that and the ethics of emergency in mind I fully support merging this PR in order to change the licensing to AGPLv4.

guilledk (approved):

goodboy:
For those interested here's some non-legal discussion on SO around this change:
https://opensource.stackexchange.com/questions/12276/how-to-change-the-license-of-a-project-from-gpl-to-agpl

overclockworked64:
I'm fine with relicensing the code.

overclockworked64 (approved):

goodboy:
Lol updated the version post haste. Not sure why 4 was in my head 😂

goodboy:
Bah I guess the commit message was wrong too 😂.
No idea why 4.

Anyway, license files and headers are all correct just commit message lul.

Being real pro at licensing details 🏄🏼, "`tractor`"

PR #276, opened 2021-12-15
Expected ctx cancelled should not override a source error

goodboy:
Fix for https://github.com/pikers/piker/issues/244 which addresses an issue where if a `Context` is cancelled due to one side erroring, we **don't want to** mask that orignal source error with the `ContextCancelled` which is responded back from the callee side which was cancelled as part of the caller side's error handling.

So in other words,
- caller opens a context with callee
- caller errors inside its context and the error handling sends a `Context.cancel()` request to the callee
- the callee side handles that cancel request and sends back a `ContextCancelled`
- the caller receives the `ContextCancelled` but should not override the local `Context._error` value with the cancel request response
- the caller should raise the original source error

### DIFF
--- a/tractor/_streaming.py
+++ b/tractor/_streaming.py
@@ -425,7 +425,6 @@ class Context:
                 f'Remote context error for {self.chan.uid}:{self.cid}:\n'
                 f'{msg["error"]["tb_str"]}'
             )
-            # await ctx._maybe_error_from_remote_msg(msg)
             self._error = unpack_error(msg, self.chan)
 
             # TODO: tempted to **not** do this by-reraising in a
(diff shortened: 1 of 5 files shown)


tractor/_streaming.py
```diff
@@ -424,8 +424,17 @@ async def _maybe_raise_from_remote_msg(
                 f'Remote context error for {self.chan.uid}:{self.cid}:\n'
                 f'{msg["error"]["tb_str"]}'
             )
-            # await ctx._maybe_error_from_remote_msg(msg)
-            self._error = unpack_error(msg, self.chan)
+            error = unpack_error(msg, self.chan)
+            if (
```
goodboy, line 428: This is the crux of the fix and should cause some of our other context streaming tests to fail currently until adjusted to more correctly expect similar source errors.

PR #279, opened 2021-12-17
Add back latest mypy since `trio-typing` 0.7.0 is now compat

goodboy:
After https://github.com/python-trio/trio-typing/issues/50 was fixed but now we have more errors from latest `mypy` 😂 

Lurkerz help!

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,7 +1,7 @@
 pytest
 pytest-trio
 pdbpp
-mypy<0.920
+mypy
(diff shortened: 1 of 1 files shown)


goodboy:
`tractor/trionics/_mngrs.py:90: error: Argument 1 to "tuple" has incompatible type "ValuesView[Optional[Any]]"; expected "Iterable[T]"`

wat.

goodboy:
Landed as part of #335.

PR #280, opened 2021-12-17
Alpha4: infect all the `asyncio`s

goodboy:
Cuz why would we go beta 🏄🏼 

Decent release here folks with lots of cool stuff 😎

The `trio`-*quarterly* summary (ping @Fuyukai):

This *quarter* in `tractor` we had our 4th alpha release with some super hip features landing as well as a ton of internal runtime improvements with the end result of a much more deterministic inter-actor cancellation system as well as better defined cross-actor task *context* semantics.

The slew of details is in [our changelog](https://github.com/goodboy/tractor/blob/master/NEWS.rst) but just in case you wanted the IDGTFTS:

- We re-licensed the code base under AGPLv3 going foward (solely to close the SaaS loophole in the old license).
- Added a new "infetced `asyncio` mode" which let's you spawn actors which run the `asyncio` event loop, start `trio` in [guest-mode](https://trio.readthedocs.io/en/stable/reference-lowlevel.html?highlight=guest%20mode#what-is-guest-mode) on the loop, then accept RPC requests from peer actors to drive `asyncio` tasks in a SC-per-task oriented way.
-  Support for *error relay* across `tractor.Context` inter-actor linked tasks meaning if a task errors in one actor the error is raised (inside an appropriate boxed exception) to the other side's cancel scope. This was part of ensuring if a stream is opened only on one side of a bi-directional stream then a new `StreamOverrun` error will be raised on the sending side to avoid unexpected backpressure hangs.
- A new  ``trionics.maybe_open_context()`` an actor-scoped async multi-task
  context manager resource caching API.
- Deterministic handling of *actor cancel requests* which are implemented currently as special RPC calls (instead of built into our messaging protocol) such that `Portal.cancel_actor()` won't not just be abandoned without waiting to see if it was successful 😂 
- Support `debug_mode: bool` on a per-actor basis.
- A bunch of bug fixes that fell out as part of landing all these features including better KBI handling in contexts, better graceful daemon-actor cancellation, fixes to the broadcast channel `trionics` implementation to correctly terminate on underlying stream closure.

Overall quite the batch of goodies 😉


goodboy:
Heh, well look at CI behaving all nice for us 😹

guilledk (approved):

goodboy:
@overclockworked64 you want a look or no?

overclockworked64 (approved):

PR #282, opened 2021-12-20
Lengthen win CI run to 12m

goodboy:
Since it's rekting our run..


goodboy:
Yeah think that was it; the 9m limit was just a bit too tight.

PR #284, opened 2021-12-22
Cancel the `.cancel_actor()` request on proc death

goodboy:
Adjust the `soft_wait()` strategy to avoid sending needless cancel
requests if suitable.

Basically an even more graceful handling of actor cancellation for parent-child peers.

### DIFF
--- a/tractor/_spawn.py
+++ b/tractor/_spawn.py
@@ -218,7 +218,9 @@ async def soft_wait(
     # ``trio.Process.__aexit__()`` (it tears down stdio
     # which will kill any waiting remote pdb trace).
     # This is a "soft" (cancellable) join/reap.
+    uid = portal.channel.uid
     try:
+        log.cancel(f'Soft waiting on actor:\n{uid}')
         await wait_func(proc)
     except trio.Cancelled:
         # if cancelled during a soft wait, cancel the child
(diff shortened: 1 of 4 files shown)


goodboy:
Uhhh wut. [Looks like we aren't using the `trio` spawn backend on windows in CI?!?](https://github.com/goodboy/tractor/runs/4610683269?check_suite_focus=true).

guilledk (approved):

tractor/_spawn.py
```diff
@@ -501,6 +523,11 @@ async def mp_new_proc(
         # local actor by the time we get a ref to it
         event, chan = await actor_nursery._actor.wait_for_peer(
             subactor.uid)
+
+        # XXX: monkey patch poll API to match the ``subprocess`` API..
+        # not sure why they don't expose this but kk.
+        proc.poll = proc._popen.poll  # type: ignore
```
goodboy, line 529: @richardsheridan mentioned on matrix that `proc.exitcode` is actually already this + some error checking - so maybe use that?

goodboy: Oh right, but i guess then we have to wrap reading that attr as an equivalent duck-typed `.poll()`?

goodboy:
🤞🏼 cmonnn CI

PR #288, opened 2022-01-03
Always raise a received result-as-error in spawn tasks

### LINKED ISSUE #287: Tasks are not cancelled
Hello,
I am on WSL2, with python 3.9.9 and tractor 0.1.0a4.

I am testing the cancellation of the tasks when one raises. I am using this simple snippet:
```python
import tractor
import trio
from trio import sleep
from trio import sleep_forever

async def not_raising() -> None:
    print("inside not raising")
    await sleep_forever()

async def raising() -> None:
    print("inside raising!")
    await sleep(5)
    raise ValueError("random raise from bad task")

async def main() -> None:
        t1 = await n.run_in_actor(not_raising, name="raising")
        t2 = await n.run_in_actor(raising, name="not raising")
    print("outside the tractor nursery")

async def main2() -> None:
    async with trio.open_nursery() as n:
        n.start_soon(not_raising)
        n.start_soon(raising)
    print("outside trio nursery")

if __name__ == "__main__":
    trio.run(main)
```

I would expect that when `raising` task raises `ValueError`, the other task gets cancelled (as Trio does, as in `main2` function). However, when I run this code, i get:
```
No actor could be found @ 127.0.0.1:1616
inside not raising
inside raising!
Jan 03 16:03:58 (not raising, 11045, raising)) [ERROR] tractor _actor.py:260 Actor crashed:
  File "/home/didi/leaf/leaf-radio/.venv/lib/python3.9/site-packages/tractor/_actor.py", line 232, in _invoke
    await chan.send({'return': await coro, 'cid': cid})
  File "/home/didi/leaf/leaf-radio/gse/main.py", line 27, in raising
    raise ValueError("random raise from bad task")
ValueError: random raise from bad task
Remote context error for ('not raising', '8e3309a2-2e39-4a8e-a3ec-15ce0977c1a8'):244eb763-1b98-4f46-9d09-93aa6791e400:
  File "/home/didi/leaf/leaf-radio/.venv/lib/python3.9/site-packages/tractor/_actor.py", line 232, in _invoke
    await chan.send({'return': await coro, 'cid': cid})
  File "/home/didi/leaf/leaf-radio/gse/main.py", line 27, in raising
    raise ValueError("random raise from bad task")
ValueError: random raise from bad task

Cancelling ('not raising', '8e3309a2-2e39-4a8e-a3ec-15ce0977c1a8') after error ('not raising', '8e3309a2-2e39-4a8e-a3ec-15ce0977c1a8')
  File "/home/didi/leaf/leaf-radio/.venv/lib/python3.9/site-packages/tractor/_actor.py", line 232, in _invoke
    await chan.send({'return': await coro, 'cid': cid})
  File "/home/didi/leaf/leaf-radio/gse/main.py", line 27, in raising
    raise ValueError("random raise from bad task")
ValueError: random raise from bad task

Cancelling existing result waiter task for ('not raising', '8e3309a2-2e39-4a8e-a3ec-15ce0977c1a8')
```
and the programs hangs indefinitely. I need to press `ctrl+c` to kill the process.

Since this is a basic feature of the library, i feel like i am missing something. 
Could you please help me understand what it is happening?

goodboy:
Hopefully fixes #287.

This small change ensures that any actor spawn task that receives an error from its remote target will raise that error locally thus triggering a surrounding nursery error and possible teardown.

ping @didimelli

### DIFF
--- a/tractor/_spawn.py
+++ b/tractor/_spawn.py
@@ -57,7 +57,7 @@ from .log import get_logger
 from ._portal import Portal
 from ._actor import Actor
 from ._entry import _mp_main
-from ._exceptions import ActorFailure
+from ._exceptions import ActorFailure, NoResult
 
 
 log = get_logger('tractor')
@@ -136,7 +136,10 @@ async def exhaust_portal(
         # always be established and shutdown using a context manager api
         final = await portal.result()
 
-    except (Exception, trio.MultiError) as err:
+    except (
+        Exception,
+        trio.MultiError
+    ) as err:
         # we reraise in the parent task via a ``trio.MultiError``
         return err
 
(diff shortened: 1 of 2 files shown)


goodboy:
Ahah, so the (main) issue with this solution is that we don't get all errors from all child actors to be accumulated as previous since some actors may get cancelled and their results (even when errors) abandoned.

I've been trying some other things but likely the way to solve the underlying issue **and** get as much error collecting as possible would be to do a 2nd pass of all child portals at nursery close once all subprocs are known to be terminated and then fill any later received errors into the `errors: dict`.

goodboy:
Oh yeah ignore the `mp` runs there 😂 
Code hasn't been changed for that backend!

tractor/_spawn.py
```diff
@@ -369,6 +397,11 @@ async def new_proc(
                     f"{subactor.uid}")
                 nursery.cancel_scope.cancel()
 
+                # if errors:
```
goodboy, line 400: This was the alternative approach mentioned in #287: cancelling the `ria_nursery` instead of raising any final error retrieved from the portal.

I think this approach is a bit more, hard to extend wrt supervisor strats being overloaded by a user.

goodboy:
> close rationale authored by `opencode`

Closing as superseded by PRs #481 and #484.

This branch tried to surface `run_in_actor()` remote errors immediately
from actor-nursery spawn/result-waiter tasks, including a second scan for
late remote errors after process exit. The discussion on issue #287
correctly identified that result and error ownership was sitting at the
wrong structured-concurrency layer.

PR #481 introduced `tractor.to_actor.run()`, which waits for and raises a
remote result or error directly in the caller's task. PR #484 then removed
`ActorNursery.run_in_actor()`, `Portal.result()`, `Portal.wait_for_result()`,
the secondary one-shot nursery, and the result-reaping machinery modified
by this branch.

The user-visible requirement now has direct coverage in
`tests/test_to_actor.py`, while this patch targets code paths that no
longer exist. Its more permissive late-error collection was intentionally
replaced by normal caller-scoped Trio cancellation semantics.

Issue #294, opened 2022-01-27
Toying with a message type for object references


goodboy:
As a future requirement and probably large convenience for many users looking to lookup and use *remote python* objects in actors i've started toying with some ideas for making a built-in message-serialization compatible type that we can offer as part of our public sugar apis.

The main set of tools include
- `msgpack` type extensions (since that's our only official codec atm):
  - https://jcristharif.com/msgspec/extending.html
    - alternatively we could go full struct style https://jcristharif.com/msgspec/api.html#struct
  -  https://github.com/msgpack/msgpack-python#packingunpacking-of-custom-data-type
- the new [pkgutils.resolve_name()`](https://docs.python.org/3/library/pkgutil.html#pkgutil.resolve_name) resolver in py 3.9

I have a draft implementation for a `str` derivative which gets us a nice compact api for conversion between `str` and `object` references and could likely be integrated with our `enable_modules:list[str]` "module capability per actor" system to get us moving in the cap-based-sec direction.

I'll put up a matching PR shortly 😎

goodboy:
Also added a couple [todo lists in the module](https://github.com/goodboy/tractor/blob/master/tractor/msg.py#L36) for any aspiring *messageers* to have a go at 😉

Issue #196, opened 2021-02-24
Typed messaging and validation


goodboy:
I *was* originally going to make a big post on [`pydantic`](https://pydantic-docs.helpmanual.io/) and how we could accomplish this typed messages using that very very nice project despite there being a couple holdups for integration with `msgpack`. 

However, it turns out just today an even faster and msgpack specific project was released: [`msgspec`](https://pydantic-docs.helpmanual.io/) 🏄🏼 

It claims to not only be faster then [msgpack-python](https://github.com/msgpack/msgpack-python) but also supports [schema evolution](https://jcristharif.com/msgspec/#schema-evolution) and other [niceties](https://jcristharif.com/msgspec/#highlights)
It also has perf bumps when making multiple [repeated encode/decode calls](https://jcristharif.com/msgspec/#usage) which is exactly [how we're currently using `msgpack` inside our `Channel`](https://github.com/goodboy/tractor/blob/master/tractor/_ipc.py#L45).

Overall there looks to be no downside and we'll get typed message semantics fast and free 👍🏼 

For reference, I'll leave a bunch of links I'd gathered regarding making `pydantic` work with `msgpack`:
- https://github.com/samuelcolvin/pydantic/issues/951
- https://pydantic-docs.helpmanual.io/usage/dataclasses/
- https://github.com/samuelcolvin/pydantic/pull/595
- https://github.com/tiangolo/fastapi/issues/1285
- https://github.com/MolSSI/QCElemental/blob/master/qcelemental/models/basemodels.py#L121

ryanhiebert:
That's really neat! I was looking at implementing Pydantic in a project a little while ago, and chose not to. It seemed like the API wasn't quite what I was looking for. I was wanting data classes, and confidence that serialization and deserialization were both strict. I'm not quite sure why I concluded that, unfortunately. I knew about the data classes integration with Pydantic, but there was something missing with it that I felt I needed.

msgspec looks pretty cool for when you control the data format, but that definitely wasn't part of what I was doing. (I was writing an API wrapper over a JSON API).

I know many people have gotten a lot of mileage out of Pydantic. It's a great project.

goodboy:
Yeah alternatively we've been thinking about using [`capnproto`](https://capnproto.org/rpc.html) and in particular seeing if we can auto-gen [schema](https://capnproto.org/language.html#language-reference) from type annotated Python functions.

I think this would be a huge boon since we'd get CBS (capability based sec) for free 🏄🏼.

[...]

PR #300, opened 2022-02-09
Use lists by default like `msgspec`, update to latest `msgspec`  and `msgpack` releases

goodboy:
As per https://github.com/jcrist/msgspec/issues/30#issuecomment-869084723 there's not much of a benefit in our use case (IPC messaging often with dynamic lists) and we can always (and are planning to) support native structs eventually anyway.

The further benefit is that we can assume `msgspec` and `msgpack` will be uniform in the array format after decode. 

Relates to https://github.com/goodboy/tractor/issues/196 and https://github.com/goodboy/tractor/issues/294.


goodboy:
Lul [`msgpec` version update](https://jcristharif.com/msgspec/usage.html#encoding) broke everything.

I'll get to it 🏄🏼

guilledk (approved):

PR #304, opened 2022-02-24
`LinkedTaskChannel.subscribe()`, explicit `asyncio` task cancel logging, `test_trioisms.py` 

goodboy:
Basically 2 parts:
- some better error logging and a somewhat involved scenario test that includes use of,
  - a sub-actor managed "service nursery" that can be used to service new async requests from a parent actor
  - broadcast channel usage for by multiple such "service tasks" within the sub-actor
  - ensuring errors in any service task cancels the others
  - making sure it all works with *infected asyncio` mode
- a new `to_asyncio.LinkedTaskChannel.subscribe()` which gives task-oriented broadcast functionality semantically equivalent to `MsgStream.subscribe()`
  - this makes it possible for multiple `trio`-side tasks to consume `asyncio`-side task msgs in tandem

This work was originally inspired as part of debugging a case in `piker`'s service manager that ended up leading to https://github.com/python-trio/trio/issues/2258.

Included is a test that demonstrates the current `trio` behaviour in a new `test_trioisms.py` for tracking such situations in the future.

---
### TODO:
- [x] at least one test for the new `LinkedTaskChannel.subscribe()` support
- [x] nooz file


goodboy:
Lul no idea but now windows wants to pass, k.

PR #307, opened 2022-04-12
Sort `.subscribe()` results before comparison

goodboy:
Post-op from #304 just to get the test more reliable since windows has order issues apparently likely due to slight scheduling/latency differences..


goodboy:
Replaced by #308.

Issue #293, opened 2022-01-26
PyPI packaging broken for version 0.1.0a4; missing README.rst


iamzoltan:
pip install fails and auto reverts to version 0.1.0a3.

this is likely due to a change in the location of the README; from the root dir to docs/.

a possible solution to point to the new location is to try using this flag in the setup.py file:

```
setup(
    ...
    include_package_data=True
)
```

source: https://setuptools.pypa.io/en/latest/userguide/datafiles.html

goodboy:
@iamzoltan thanks for the issue friend!

Yeah we clearly dropped a couple balls on this readme move 😂

Is there a way we can test this without using a `pypi` server?

goodboy:
i guess the easiest is to just test on the test pypi server, heh,

Will try to get to this today but either way i think the last (2?) alphas are borked from this yah?

PR #312, opened 2022-07-10
Include ./docs/README.rst in src dist

goodboy:
Fixes #293 hopefully 😂

PR #314, opened 2022-07-11
Add an sdist install job

goodboy:
This should hopefully catch issues like,
https://github.com/goodboy/tractor/issues/293


goodboy:
Pulled in commit from #312 to demonstrate it fixes #293 😎

PR #316, opened 2022-07-12
Try windows CI on py 3.10

goodboy:
Like it sounds.


goodboy:
Also made https://github.com/pdbpp/pdbpp/issues/498 to track this since it's the reason we're seeing it.

PR #322, opened 2022-08-03
3.10 and friends

goodboy:
Official move to strictly supporting python 3.10+ and beginning a "core runtime" internal code rework to look more like ``trio``'s organization.

The main motivation for this change is to begin use of the new `match:` syntax in our internal RPC msg loops 😎 

---
### high level reorg
- rename `._actor.py` -> `_runtime.py`
- move a bunch of rather large singleton-task-executed runtime methods to be module level functions [to look more like the internals of `trio` itself](https://github.com/python-trio/trio/blob/master/trio/_core/_run.py#L1852). namely the following routines:
  - `Actor._process_messages()` and `._async_main()`
  
---
#### Landing ToDo:
- [x] nooz file
- [x] readme update mentioning windows support but current state of no active testing/CI **=> moved to #327**..
- [x] drop `tractor.run()` once and for all - https://github.com/goodboy/tractor/issues/183
- [x] drop lingering uses of old `typing.Dict`, `Tuple`, `List` annotations


tractor/_debug.py
```diff
@@ -126,7 +126,8 @@ def release(cls):
         try:
             # sometimes the ``trio`` might already be terminated in
             # which case this call will raise.
-            cls.local_pdb_complete.set()
+            if cls.local_pdb_complete is not None:
```
goodboy, line 129: This is history from https://github.com/goodboy/tractor/pull/324 which we can probably land first.

guilledk (approved):

PR #325, opened 2022-08-24
Make `maybe_open_context()` re-entrant safe, use per factory locks

goodboy:
As discovered and discussed in code review of https://github.com/pikers/piker/pull/331 during refactoring we can't currently support re-entrant usage of `tractor.trionics.maybe_open_context()`.

This fixes that by using a per-target `id(acm)` lock table to avoid deadlock on the previous single global lock design.

---
TODO:
- [ ] tests..

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -396,7 +396,7 @@ tasks spawned via multiple RPC calls to an actor can modify
 
 
         # a per process cache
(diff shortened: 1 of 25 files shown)


guilledk (approved):
Tested on `deribit` branch of `piker`, good work.

goodboy:
Lol what's with windows CI all of sudden?

PR #329, opened 2022-10-06
Nursery per `trionics.maybe_open_context()`

goodboy:
So avoid using the actor's global service nursery and instead allocate local nurseries per manager-`def`-instance set.

This also adds `key: Callable[..., Hashable]` support and moves `maybe_open_nursery()` into this `trionics._mngrs.py` module. 

Needs to be rebased to master obviously.


goodboy:
In favor of  #336

PR #331, opened 2022-10-07
Drop reporting flag from testsuite

goodboy:
attempting to avoid a weird crash in CI seen in #329


goodboy:
wasn't the source issue for CI failure.

Issue #332, opened 2022-10-08
`trio.MultiError` is deprecated -> move to `exceptiongroup.ExceptionGroup`


goodboy:
This is killing CI: https://github.com/goodboy/tractor/actions/runs/3207101811/jobs/5241628100

```python
test_multierror_assert.py::test_bad_assert
  /home/goodboy/repos/tractor/test_multierror_assert.py:32: TrioDeprecationWarning: trio.NonBaseMultiError is deprecated since Trio 0.22.0; use ExceptionGroup (on Python 3.11 and later) or exceptiongroup.ExceptionGroup (earlier versions) instead (https://github.com/python-trio/trio/issues/2211)
    assert trio.NonBaseMultiError is trio.MultiError
```

Have repro-ed this locally

goodboy:
Closed by #333

PR #333, opened 2022-10-09
`ExceptiongGroup`s and `trio>=0.22`

goodboy:
Attempting to get the new `py3.11` driven [exception group](https://peps.python.org/pep-0654/#exceptiongroup-and-baseexceptiongroup) backport support in. See [pep 654](https://peps.python.org/pep-0654/#handling-exception-groups) for dirty deats.

Pertains to fixing #332 and appears to have caused a bunch of test hangs particularly in the debugger suite (which I've manually verified) which need some investigation.

**UPDATE** tests now fixed with the landing of #337 🏄🏼

### DIFF
--- a/examples/debugging/multi_daemon_subactors.py
+++ b/examples/debugging/multi_daemon_subactors.py
@@ -27,7 +27,18 @@ async def main():
 
         # retreive results
         async with p0.open_stream_from(breakpoint_forever) as stream:
-            await p1.run(name_error)
+
+            # triggers the first name error
+            try:
+                await p1.run(name_error)
+            except tractor.RemoteActorError as rae:
+                assert rae.type is NameError
+
+            async for i in stream:
+
+                # a second time try the failing subactor and this tie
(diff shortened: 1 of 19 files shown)


goodboy:
Beauty, now it's just some debugger test cases which are borked..

I've got a patch which solves this (as well as a general solution for the "un-contactable subactor left a stale lock state") but, it's going to require these the tests to be adjusted since they'll fail earlier now with less locks due to the new "block list" approach working.

goodboy:
Hmm i might break out the new debugger fix stuff into a new dependent PR since it is somewhat unrelated to this specific change set.

Also might do the same for the uuid-in-log msgs adjustment?

goodboy:
Alright factored out the relevant debugger locking solution stuff into #337.

guilledk (approved):

goodboy:
💥 almost there!

goodboy:
All the mega-deats in the nooz file 🏄🏼

PR #337, opened 2022-10-12
Debug lock blocking

goodboy:
Pulled out from #333.

Provides a (decent) solution to the problem of a sub-actor (of the root) who loses IPC connectivity but still tries to acquire the stdstreams / TTY lock. This failure mode can result in a deadlock where the root thinks the child is in debug while the child is waiting to be cancelled and because of our debug mode `SIGINT` override handler, the root and child never tear down on `ctl-c` from the user..

---
This patch solves that by taking a simple approach: whenever the root actor detects that some peer (who presumably is some child-ish) has no more working IPC (channel) connections, will stick the sub's `Actor.uid` on a new `._debug.Lock._blocked` list such that the actor is furthermore blocked from acquiring the lock and will be rejected from such with a `ContextCancelled`.

---
Additionally included here is some test fixes which accommodate this change, mainly some of the debugger tests terminating earlier then previous due to the block list preventing recurrent locks from a nested child who's parent has already been cancelled and thus IPC disconnected.

Also, is a couple tweaks to the `asyncio` test suite discovered in the #333 tinkerings.

### DIFF
--- a/tests/test_infected_asyncio.py
+++ b/tests/test_infected_asyncio.py
@@ -203,24 +203,25 @@ def test_context_spawns_aio_task_that_errors(
     '''
     async def main():
 
-        async with tractor.open_nursery() as n:
-            p = await n.start_actor(
-                'aio_daemon',
-                enable_modules=[__name__],
-                infect_asyncio=True,
-                # debug_mode=True,
-                loglevel='cancel',
-            )
-            async with p.open_context(
-                trio_ctx,
-            ) as (ctx, first):
+        with trio.fail_after(2):
+            async with tractor.open_nursery() as n:
+                p = await n.start_actor(
+                    'aio_daemon',
+                    enable_modules=[__name__],
+                    infect_asyncio=True,
+                    # debug_mode=True,
+                    loglevel='cancel',
+                )
+                async with p.open_context(
+                    trio_ctx,
+                ) as (ctx, first):
 
-                assert first == 'start'
+                    assert first == 'start'
 
-                if parent_cancels:
-                    await p.cancel_actor()
+                    if parent_cancels:
+                        await p.cancel_actor()
 
-                await trio.sleep_forever()
+                    await trio.sleep_forever()
 
     with pytest.raises(RemoteActorError) as excinfo:
         trio.run(main)
--- a/tests/test_infected_asyncio.py
+++ b/tests/test_infected_asyncio.py
@@ -170,11 +170,11 @@ async def trio_ctx(
     # message.
     with trio.fail_after(2):
         async with (
+            trio.open_nursery() as n,
+
             tractor.to_asyncio.open_channel_from(
                 sleep_and_err,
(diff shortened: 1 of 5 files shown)


guilledk (approved):

tests/test_infected_asyncio.py
```diff
@@ -170,11 +170,11 @@ async def trio_ctx(
     # message.
     with trio.fail_after(2):
         async with (
+            trio.open_nursery() as n,
```
goodboy, line 173: Just making a mental note of this change: obviously it makes sense for this test to hang if the cancellation from the `.open_channel_from()` never propagates bc the nursery was still sleeping.

tests/test_infected_asyncio.py
```diff
@@ -203,24 +203,25 @@ def test_context_spawns_aio_task_that_errors(
...
-                'aio_daemon',
-                enable_modules=[__name__],
-                infect_asyncio=True,
-                # debug_mode=True,
-                loglevel='cancel',
-            )
-            async with p.open_context(
-                trio_ctx,
-            ) as (ctx, first):
+        with trio.fail_after(2):
```
goodboy, line 206: Avoids this test ever hanging again..

tractor/_debug.py
```diff
@@ -72,6 +78,18 @@ class Lock:
...
+    # the CALLER side of the `lock_tty_for_child()` context entry-call
+    # and must be cancelled if this actor is cancelled via IPC
+    # request-message otherwise deadlocks with the parent actor may
+    # ensure
+    _debugger_request_cs: Optional[trio.CancelScope] = None
+
+    # NOTE: set only in the root actor for the **local** root spawned task
+    # which has acquired the lock (i.e. this is on the callee side of
+    # the `lock_tty_for_child()` context entry).
+    _root_local_task_cs_in_debug: Optional[trio.CancelScope] = None
```
goodboy, line 91: using this seems to cause a lot of test failures and *too-early-cancellation-of-the-child(s)* so keeping the code there but not currently using it inside the `._runtime` changes for now.

tractor/_runtime.py
```diff
@@ -702,6 +712,34 @@ async def _stream_handler(
...
+                    # we cancel it to allow servicing other requests for
+                    # the lock.
+                    if (
+                        pdb_lock._root_local_task_cs_in_debug
+                        and not pdb_lock._root_local_task_cs_in_debug.cancel_called
+                    ):
+                        log.warning(
+                            f'STALE DEBUG LOCK DETECTED FOR {uid}'
+                        )
+                        # TODO: figure out why this breaks tests..
```
goodboy, line 740: As mentioned above, so far it seems just putting the child on the block list is enough and avoids cancelling some actor that for wtv reason became IPC disconnected but still is using the debugger. My guess is there's a race between the scheduling of this cancel and a new request coming in?

Needs more tinkering/testing.

PR #344, opened 2022-12-12
Harden cluster tests

goodboy:
In response to a weird issue discovered by @guilledk where if you pass in a `mngrs=(<generator comprehension>)`, you end up with an empty sequence in the loop inside `trionics.gather_contexts()` and then  a hang due to waiting on the `all_entered: trio.Event` [just after the loop](https://github.com/goodboy/tractor/blob/master/tractor/trionics/_mngrs.py#L123).

This solves the issue by greedly converting the `mngrs` sequence to a list and ensuring its size and otherwise raising a `ValueError`.

---
Further TODO for this to land:
- [x] demonstrate issue through hanging test no commit for the fix yet pushed
  - exact test failure [here in CI run](https://github.com/goodboy/tractor/actions/runs/3671732526/jobs/6207227497#step:6:18) from commit [c606be8](https://github.com/goodboy/tractor/pull/344/commits/c606be8c640d43da260ac1d457a3551150b3d85e)
- [x] push the actual fix ([b5192cc](https://github.com/goodboy/tractor/pull/344/commits/b5192cca8ef32d89ab46ba98d231bb47aa1218a3))
- ~add tests for both the generator and empty list input cases~ there's no real reason for this, and further trying to jam in the logic in a `pytest` parametrization is janky..
- [x] nooz snippet

---
Other additions part of this patch:
- we now passthrough all runtime kwargs to the `open_actor_cluster()` helper such that any kwarg to `open_root_actor()` can be used

### DIFF
--- a/nooz/344.bugfix.rst
+++ b/nooz/344.bugfix.rst
@@ -1,0 +1,11 @@
+Always ``list``-cast the ``mngrs`` input to
+``.trionics.gather_contexts()`` and ensure its size otherwise raise
+a ``ValueError``.
+
+Turns out that trying to pass an inline-style generator comprehension
+doesn't seem to work inside the ``async with`` expression? Further, in
+such a case we can get a hang waiting on the all-entered event
+completion when the internal mngrs iteration is a noop. Instead we
+always greedily check a size and error on empty input; the lazy
+iteration of a generator input is not beneficial anyway since we're
+entering all manager instances in concurrent tasks.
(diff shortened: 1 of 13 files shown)


guilledk (approved):

goodboy:
interesting i think [38326e8](https://github.com/goodboy/tractor/pull/344/commits/38326e8c155052b5b60cc29b730dad0b61d2ed73) makes the debugger tests pass that were failing on last run. I did a little local manual testing and it seems there can be transient failures on teardown..

I'm not sure if we should write an independent test to catch this since it's more of an *auxillary* teardown error that shouldn't (often) be a direct issue?

Issue #345, opened 2022-12-12
`OSError: [Errno 9] Bad file descriptor` in cluster test


goodboy:
Due to some change in the downstream history introduced in #343.

Thanks to @guilledk who also discovered this issue in practice during dev of a client project. 

Not entirely sure which commit(s) are causing it yet but the error looks as follows (pulled out from #344):

 ```python
 ERROR    tractor._supervise:_supervise.py:385 Nursery for ('root', '798cdc51-2c9e-4041-a162-6c4d0334feed') errored with
  + Exception Group Traceback (most recent call last):
  |   File "/home/goodboy/repos/tractor/tractor/_supervise.py", line 339, in _open_and_supervise_one_cancels_all_nursery
  |     yield anursery
  |   File "/home/goodboy/repos/tractor/tractor/_supervise.py", line 499, in open_nursery
  |     yield anursery
  |   File "/home/goodboy/repos/tractor/tractor/_clustering.py", line 58, in open_actor_cluster
  |     async with trio.open_nursery() as n:
  |   File "/home/goodboy/.virtualenvs/310tractor/lib/python3.10/site-packages/trio/_core/_run.py", line 850, in __aexit__
  |     raise combined_error_from_nursery
  | trio.NonBaseMultiError: OSError(9, 'Bad file descriptor'), OSError(9, 'Bad file descriptor'), OSError(9, 'Bad file descriptor'), OSError(9, 'Bad file descriptor')
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "/home/goodboy/repos/tractor/tractor/_clustering.py", line 63, in _start
    |     portals[name] = await an.start_actor(
    |   File "/home/goodboy/repos/tractor/tractor/_supervise.py", line 155, in start_actor
    |     parent_addr = self._actor.accept_addr
    |   File "/home/goodboy/repos/tractor/tractor/_runtime.py", line 1175, in accept_addr
    |     return self._listeners[0].socket.getsockname()  # type: ignore
    | OSError: [Errno 9] Bad file descriptor
    +---------------- 2 ----------------
    | Traceback (most recent call last):
    |   File "/home/goodboy/repos/tractor/tractor/_clustering.py", line 63, in _start
    |     portals[name] = await an.start_actor(
    |   File "/home/goodboy/repos/tractor/tractor/_supervise.py", line 155, in start_actor
    |     parent_addr = self._actor.accept_addr
[...]

PR #346, opened 2022-12-12
Ipc failure while streaming

goodboy:
Variety of streaming related teardown fixes for when IPC goes down before streams are terminated gracefully. Mostly discovered during dev of initial draft of https://github.com/pikers/piker/pull/420.

---
#### TODO:
- [x] solve issue in this history causing #345 => **dropping [5dc07f0](https://github.com/goodboy/tractor/pull/346/commits/5dc07f04557932322895c54eeafc960217454e37) did it**
- [x] add more specific tests (particularly for b2971733c6) to match each of these "fixes" 😂


goodboy:
Lol.. ok so its [5dc07f0](https://github.com/goodboy/tractor/pull/346/commits/5dc07f04557932322895c54eeafc960217454e37)..

No idea why yet, but dropping that commit for this history and gonna rebase all dependents.

goodboy:
Lul, and turns out [ba6a2b1](https://github.com/goodboy/tractor/pull/346/commits/ba6a2b16c5fe041b41afc7677326cea36022e302)  causes the `test_context_streaming_semtantics` suite to hang 😂 

So gonna drop that one for now as well..

tractor/_runtime.py
```diff
@@ -319,7 +319,7 @@ async def _invoke(
             BrokenPipeError,
         ):
             # if we can't propagate the error that's a big boo boo
-            log.error(
+            log.exception(
```
goodboy, line 322: This is important to show *what* traceback wasn't able to be propagated to the far end actor's caller task.

tractor/_runtime.py
```diff
@@ -826,7 +829,12 @@ async def _push_result(
 
             if ctx._backpressure:
                 log.warning(text)
-                await send_chan.send(msg)
+                try:
+                    await send_chan.send(msg)
```
goodboy, line 833: If using backpressure on  a stream we need to be sure we don't crash when the local feed memchan has been broken.

Need a test for this, though I can't remember if this was the original issue that caused the `ipc_failure_during_stream.py` example to inf hang before?

tractor/_streaming.py
```diff
@@ -609,7 +610,14 @@ async def open_stream(
                 # XXX: Make the stream "one-shot use".  On exit, signal
                 # ``trio.EndOfChannel``/``StopAsyncIteration`` to the
                 # far end.
-                await self.send_stop()
+                try:
+                    await self.send_stop()
```
goodboy, line 614: Also need a test for this:

- one end of a stream is closed 
- the other end trying to close before any other stream methods `.receive()`/`.send()` are called (thus detecting the closure before trying to send a stop msg)

tractor/trionics/_mngrs.py
```diff
@@ -133,13 +133,13 @@ async def gather_contexts(
...
-
-        # we don't need a try/finally since cancellation will be triggered
-        # by the surrounding nursery on error.
-        parent_exit.set()
+        try:
+            yield tuple(unwrapped.values())
+        finally:
+            # NOTE: this is ABSOLUTELY REQUIRED to avoid
+            # the following wacky bug:
+            # <tractorbugurlhere>
```
goodboy, line 141: I don't think I can remember the case that caused this unfortunately 😂 though pretty sure it was in `piker`.

Unfortunately this may just go in without a false positive case ..

guilledk (approved):

PR #351, opened 2023-03-07
Proper `breakpoint()` hooking

goodboy:
We were being super sloppy previous and leaking our override into Python's runtime 😂 
This would result in a `NoRuntime` being raised on bp usage after runtime exit..

Again, much of this work needs final polish and refinement before we propose something more formal for https://github.com/python-trio/trio/issues/1155

This repairs that and will eventually come with at least one test to ensure everything works as expected outside `trio`/`tractor`, when the actor stack tears down.

---
#### Testing todo:
- [ ] ensure std `breakpoint()` calls work *outside*  `tractor.open_nursery()`
  - [ ] we might want to ensure something something with threads and/or maybe `numba`?
- [ ] add a test suite for use inside an infected asyncio actor:
  - [ ] `asyncio` task uses `breakpoint()` directly, expect normal UX
  - [ ] `asyncio` task crashes and trio side enters debug? (or should the aio task get caught?)
  - [ ] cancel actor from parent and ensure that if some asyncio-task has debugger lock, normal teardown blocking is adhered?


goodboy:
Replaced by #362

PR #355, opened 2023-03-31
fixed the `Zombie` example having wrong indentation

kehrazy:


goodboy (approved):
Lol weird. Nice catch and thanks!

PR #357, opened 2023-04-14
`Context` semantics for cross-actor-task cancellation and overruns

goodboy:
Finally, a rigorous rework of our IPC `Context` for strict remote
cancellation semantics via a new `ContextCancelled` error type
[to be more like how a `trio.CancelScope`](https://trio.readthedocs.io/en/stable/reference-core.html#cancellation-semantics)
handles `trio.Cancelled` but in this case expressed cross-actor-SC-`trio.Task`s !

There's **ALOT** of details that go into this patch-set and I don't
expect everything here to be groked in entirety by anyone (nor do
i expect everything written in this description to be perfectly
comprehensive with the commit msgs).

This PR is primarily delivering the initial basis for our runtime's
solution to "SC-across-processes"; it **will be followed up very
shortly** with a slew of of refinement PRs which more rigorously
define an inter-`Actor`-`trio.Task`-supervision protocol via
`msgspec`!

As such i would not consider this PR to leave the runtime in anything
more then a *highly experimental state*

BUT, after the follow PRs have landed, `tractor`'s distributed
runtime is what i'd like to dub as being "very close to being
stable"..

---

#### High level mechanics:
possible `parent -> child` msg-sequence flows now include..

- graceful cancellations:
  - parent cancels child, child cancel-acks, raises ctxc (`ContextCancelled`)
    locally.
  - child self cancels, parent raises ctxc.

  - if any actor-task requests a remote actor to be cancelled, either by
    calling `Actor.cancel()` or the internal `Actor._cancel_stask()` (which
    is called by `Context.cancel()`), that requesting actor's
    uid is set on the target far-end `Context._cancel_called_remote` and
    any eventually raised `ContextCancelled.canceller: tuple`.

  - this allows us to support so called **graceful remote cancellation**:
    - i.e any actor receiving a `ContextCancelled` error-msg can check
      if *they* were the "requester" of the remote peer task's
      cancellation.
    - in this case the error is **not** raised locally and instead the
      remote-cancellation (represented as a locally raised
      `ContextCancelled` on the requester side) is "caught" and
      silently absorbed inside the `Portal.open_context()` block (or
      alternatively in the child-side `@tractor.context` callee func)
      just like normal `trio` cancel scopes: if you call `cs.cancel()`
      inside some scope or nursery block.

  - the **NON-graceful-cancellation** case can then be defined as the inverse condition,
    - if any received `ContextCancelled.canceller != current_actor().uid`,
    - in this case the error **is raised** thus making it explicit that
      the cancellation was not anticipated by the ctx-cancelled-receiving
      (actor's) task and thus can be subsequently handled as an
      (unanticipated and possibly *cascaded*) remote cancellation.

- error condition(s):
  - parent errors, sends ctxc to child, child ctxc acks after
    cancellation.
  - child errors and relays to parent, parent raises
    `RemoteActorError` boxing traceback from child.

- streaming semantics:
  - parent opens `MsgStream`, child does not (and vice versa) where
    every mismatched case now raises pedantic `StreamOverrun`s that
    explain the un-cooperative side.
  - parent & child open streams and either side can terminate with
    a `{'stop'}` msg.
  - ctxc requested or raised during streaming session is handled
    loudly by the other side.

---

#### Various (high level) deats of the new implementation:

- move all `Context` related code to a new `._context` mod.
  - add a new `mk_context() -> Context` factory.
  - extensions to `Context`:
    - add `Context._deliver_msg()` as the factored content from what was
      `Actor._push_result()` which, now the latter calls the former.
    - add `.wait_for_result()` over `.result()` which now internally
      delegates to a `._has_final_result_set()`.
    - expose new remote cancellation properties,
      - `Context.cancel_called`,
      - `.cancelled_caught`,
      - `.cancel_acked`,
      - `.cancel_called_remote`
    - much extended `.__repr__/__str__()`-output and utilsin terms
      of,
      - child task `.outcome` via `.repr_outcome()`,
      - underlying IPC (transport) addr info via `.dst_maddr: str`
      - `.side: str` for child or parent 
      - `.repr_rpc` for `.outcome`-as-RPC-function-call repr.
    - improved internal traceback frame hiding (such that user code
      doesn't always see low-level frames).
    - absorb `trio.EndOfChannel`s raised from `.open_context()` since
      stream errors needs not propagate outside their parent's scope.

- add `ContextCancelled.canceller: tuple[str, str]` for tracing ctx
  cancel causers.

---
#### Other (related) new runtime feats:
Besides an extremely in-depth set of new test suites for all of the
above functionality there are further additions, refinements,
factorings and bug fixes; the high level summary includes:

---
##### `MsgStream` overrun tolerance via new `allow_overruns: bool`
Included is a public API for allowing contexts (and their streams) to
"allow overruns" (since backpressure isn't really possible without
a msg/transport protocol extension): enabling the case where some
sender is pushing msgs faster then the other side is receiving them
and no error is received on the sender side.

Normally (and by default) the rx side will receive (via
`RemoteActorError` msg) and subsequently raise a `StreamOverrun` that's
been relayed from the sender. The receiver then can decide how to the
handle the condition; previously any sent msg delivered that caused an
overrun **would be discarded**.

Instead we now offer a `allow_overruns: bool` to the
`Context.open_stream()` API which adds a new mode with the following
functionality:
- if an overrun condition is hit, we spawn a single "overflow drainer"
  task in the `Context._scope_nursery: trio.Nursery` which runs
  `Context._drain_overflows()` (not convinced of naming yet btw :joy:):
- this task has one simple job: it does the work of calling `await
  self._send_chan(msg)` for every queued up overrun-condition msg
  (stored in the new `Context._overflow_q: deque`) in a blocking
  fashion **but without blocking the RPC msg loop**
- when the `._in_overrun: bool` is already set we presume the task is
  already running and instead new overflow msgs are queued and
  a warning log msg emitted.
- on `Context` exit any existing drainer task is cancelled and lingering
  msg discarded but reported in a warning log msg.

The API flag is now supported via,
- `Context.open_stream()`,
- `Actor.start_remote_task()`,

Test suites which verify usage,
- `test_context_stream_semantics`
- `test_clustering`
- `test_task_broadcasting`

---
##### Add a "pretty" `msgspec.Struct` extension (brought in from dependent)
- for now placed as `.msg.types.Struct`

---
##### Miscellaneous additions
- factored inter-`Actor` msg processing machinery to a new `._rpc` mod;
  previously was lobbed in with `._runtime`.
- extend test suites to include full coverage of all IPC cancel
  request and overrun handling cases.
- improvements to `tractor.trionics` primitives/APIs:
  - support delivering `None` from `gather_contexts()`.

---
##### Dependency updates:
- pin to `trio==0.24.0` and thus use of `ExceptionGroup` over
  `trio.MultiError` but still avoid strict egs being enabled by
  default.
- drop `trio_typing` since merged into core.

---
#### Summary of new or heavily modified test-suites
- test_child_manages_service_nursery.py
- test_context_stream_semantics.py
- tests/test_inter_peer_cancellation.py
- test_advanced_streaming.py
- test_advanced_faults.py

Also we moved back to a `tractor._testing` sub-pkg for test utils
instead of the wacky multi-`conftest.py` importing that was going on
and which is no longer (easily) used with `pytest`.

---

#### Follow-up tasks for future PRs:

- [ ] for the `ContextCancelled.canceller != current_actor().uid`
  case it's handy to know why some parent actor cancelled some other
  child that *you* were attached to, instead of just seeing that it
  was cancelled by someone else?
  - [ ] a src actor (task) is reported as the origin `.canceller`

- [ ] **not** use `@dataclass` for `Context` and instead go
  `msgspec.Struct`?

---

#### Appendix
- original GH history is backed up as branch: github/ctx_cancel_semantics_and_overruns_GH_BACKUP
 https://github.com/goodboy/tractor/tree/ctx_cancel_semantics_and_overruns_GH_BACKUP


goodboy:
Also note, i didn't try to make `mypy` happy whatsoever in this PR 👅

goodboy:
yeah no idea why CI is failing with this junk.. it all runs fine on py3.11 locally for me 🤷🏼‍♂️ 

```
tests/test_docs_examples.py:93: in <lambda>
    ids=lambda t: t[1],
E   TypeError: 'NotSetType' object is not subscriptable
```

only thing i can think of is that i'm currently running `pytest==8.3.4` ??

PR #358, opened 2023-04-16
Switch to `pdbp` 🏄🏼 

goodboy:
A ton of thanks to @mdmintz for the fork and fix of `pdbpp` in the new
and improved `pdbp`: https://github.com/mdmintz/pdbp

It would seem pretty much all the outstanding issues we had with `pdbpp`
are non-existent in the new project 💃🏼 , so i think other then running
this in my production dev eng for a bit, we can probably just land this
pronto 😎 

---
#### Deatz of previous issues that seem to now be resolved:
- [x] `pytest` top level root process wasn't using `pdbpp` on test
  crashes (when using `--pdbp` flag; seems completely resolved now
- [x] no need to uninstall `pyrepl` any more since the `alt-l` hotkey
  i've been wanting works just fine in vim mode?
- [x] previously we weren't disabling the `truncate_long_lines: bool`
  which solves the whole resized-terminal-messes-with-vertical-line-spacing.
- [x] current stack frame highlighting as suggested in https://github.com/mdmintz/pdbp/issues/32

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -6,8 +6,14 @@
 ``tractor`` is a `structured concurrent`_, multi-processing_ runtime
 built on trio_.
 
-Fundamentally ``tractor`` gives you parallelism via ``trio``-"*actors*":
-our nurseries_ let you spawn new Python processes which each run a ``trio``
+Fundamentally, ``tractor`` gives you parallelism via
+``trio``-"*actors*": independent Python processes (aka
+non-shared-memory threads) which maintain structured
+concurrency (SC) *end-to-end* inside a *supervision tree*.
+
+Cross-process (and thus cross-host) SC is accomplished through the
+combined use of our "actor nurseries_" and an "SC-transitive IPC
+protocol" constructed on top of multiple Pythons each running a ``trio``
 scheduled runtime - a call to ``trio.run()``.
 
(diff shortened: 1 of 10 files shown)


mdmintz:
There's a new release of ``pdbp`` - ``1.3.0`` - https://github.com/mdmintz/pdbp/releases/tag/v1.3.0

mdmintz:
``pdbp`` ``1.3.1`` is available, in case anyone missed the news.

goodboy:
@mdmintz thanks! been trying it out the last day or so.

I'm still slightly confused on the truncate defaults *i think?*
Not sure whether we want it disabled or not to keep  vertical spacing between lines always consistent?

mdmintz:
A new release of ``pdbp`` is available - https://github.com/mdmintz/pdbp/releases/tag/v1.4.0

goodboy:
Weird, debugger tests failed even though they didn't before?

Removed the `pdbpp` dep from test reqs, let's see if that does it?

goodboy:
Only last thing was maybe removing the `pyreadline3` dep from the setup since it was only required on windows IIRC?

Not sure it's a big deal enough to block this and we probably won't get clarity until @mdmintz returns from vaca anyway 😂 

So gonna land once stamped!

Issue #354, opened 2023-03-28
`infected_asyncio` task hang (`SIGINT` ignored!) on debug mode breakpointing?


goodboy:
As found in d6ed97d0 of the #351 work, there is a nasty issue that
causes an un-interrupt-able hang when we use enough
`tractor.breakpoint()`s in the right spots *and* infected `asyncio` mode
where the `asyncio` task somehow gets stuck in some kind of cancelled
state (maybe?)

The whole thing needs to be more thoroughly investigated and a proper
test suite built out.

---
ToDo testing:
- [ ] at least one test that actually uses the flags intro-ed in
  d6ed97d0 with a comparable timeout.
  - [ ] a fix for that case ^ :joy:

PR #362, opened 2023-05-15
`asyncio` (mode) debugger support

goodboy:
Replacement for #351 to start effort toward testing our debugger (with
`pdbp`) and basic functionality when running actors in
`infected_asyncio` mode.

---
#### Test set todo:
- [ ] ensure std `breakpoint()` calls work *outside*  `tractor.open_nursery()`
  - [ ] we might want to ensure something something with threads and/or maybe `numba`?
- [ ] add a test suite for use inside an infected asyncio actor:
  - [ ] `asyncio` task uses `breakpoint()` directly, expect normal UX
  - [ ] `asyncio` task crashes and trio side enters debug? (or should the aio task get caught?)
  - [ ] cancel actor from parent and ensure that if some asyncio-task has debugger lock, normal teardown blocking is adhered?

---
#### Any new issues?
- [ ] go through our list here: https://github.com/goodboy/tractor/issues?q=is%3Aissue+is%3Aopen+label%3Adebugger
  - [ ] of particular note is the hang issue recently found in
    https://github.com/goodboy/tractor/issues/354, which has an example
    / test added in #351

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -3,8 +3,8 @@
 |gh_actions|
 |docs|
 
-``tractor`` is a `structured concurrent`_, multi-processing_ runtime
-built on trio_.
+``tractor`` is a `structured concurrent`_, (optionally
+distributed_) multi-processing_ runtime built on trio_.
 
 Fundamentally, ``tractor`` gives you parallelism via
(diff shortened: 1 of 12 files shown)


goodboy:
Included in the already landed history of both #372 and #374.

Issue #22, opened 2018-07-28
Supervisors - what are they good for?


tgoodlet:
I want to start design discussion on how to approach [erlang-like supervisors](https://learnyousomeerlang.com/supervisors) including restart strategies and how this will interplay with a [service discovery](https://en.wikipedia.org/wiki/Service_discovery) system and possibly [real-time code replacement](http://erlang.org/doc/design_principles/release_handling.html#11).

Some questions/comments I have:
- what's the (semantic) difference between a `nursery` and a `supervisor`?
  - a nursery is more primitive and is what a supervisor should be built upon (eg. python-trio/trio#569)?
  - should a `supervisor` be specific to managing (possibly distributed) processes in which case does having a `nursery` mean anything other then a special *type* of supervisor?
  - there's been some interesting talk of [renaming nurseries](https://github.com/python-trio/trio/issues/504#issuecomment-403822271) and how [`asyncio` might be approaching nurseries/supervisors](https://gitter.im/python-trio/general?at=5b584ac75301255724c3e8e2)
- an explicit distinction should be made between a `MainProcess` supervisor and the `arbiter` actor
  - the `Arbiter` is per-host and is part of *service discovery* between hosts

- how do we best implement a *distributed supervisor* (one that spawns actors over multiple hosts)?
  - does it simply send the spawn request to the appropriate `arbiter` who will then create or use a host-local nursery?
  - what happens when all actors on the remote host go down but a remote actor is still using a remotely spawned actor in that process cluster?  
  
- what does a *distributed* process supervisor API look like?

  - how does this interact with the `arbiter` and service discovery system?

- how does an orchestration layer build on all this?

Much more to come...

goodboy:
Worth a read is `elixir`'s page on [supervisors](https://elixir-lang.org/getting-started/mix-otp/supervisor-and-application.html) and [dynamic supervisors](https://elixir-lang.org/getting-started/mix-otp/dynamic-supervisor.html).

goodboy:
We had a brief discussion a while back slightly in regard to #190 related apis.

For supervisors I actually think we can make quite a nice api by just offering composed async context managers:

```python
async with tractor.open_nursery() as n:

    async with tractor.supervise.one_for_one(n) as one_for_one_nursery:
        
        await one_for_one_nursery.run_in_actor(blah)
```

This has the benefit of being simple to implement:
- error handling logic is completely self contained in the specialized - *lower* super
[...]

PR #363, opened 2023-05-17
`taskman`: an initial prototype for a `trio.Nursery`-like per-task-scope-manager (maybe via a user defined single-yield-generator function?) 😎 

goodboy:
As per some discussion in the `trio` gitter and starting something
toward #22 and supporting our debugger crash mode from within
a `trio`-compat task nursery.

---
### Idea: a `trio.Nursery` API for optionally allowing a user to define a per-task *manager*

But why?

Well if you wanted any of the following (some of which are implemented
in the module script in this patch):

- generally gain access to lower level `trio.Nursery` *"internals"* for any
  per-task oriented purpose.
- some way to wrap your own "task handle" so you could do things like
  waiting on a task to finish and get its result.
- allow hooking into each task's exit/teardown (phase) for the purposes
  of debugging crashes, (unexpected) cancels or just generally tracing
  the `trio` per task control flow.
- enable more granular task supervision strategies like `one_for_one`
  from erlang/elixir and friends.

---
#### ToDo:
- [x] offer a `task_manager` context-manager-as-generator API such
  that the current `TaskOutcome` struct **is not** tied to the
  internals of the `ScopePerTaskNursery.start_soon()` implementation.
  - [x] ideally the `outcome: Outcome = yield (cs, ..)` is delivered to
    the user defined mngr ~~via a `Outcome.send(mngr)` call~~ such that only
    raw `trio` (dependency) internal types are provided to the user for
    wrapping / adjusting.
    - **NOTE:** didn't use `Outcome.send()` bc it would send the underlying
      value, not the outcome instance; not sure if that's best?

- [ ] **way better naming** since most of this was whipped up on the
  first try and getting good names is probably going to be tricky :joy:
  - [x] `@task_manager` is maybe better? `@task_scope_manager`?
  - [x] `ScopePerTaskNursery` is terrible.. how about `TaskScopedNursery`?
    - now i'm thinking you don't need much at all, `TaskManagerNursery`?
  - [x] `TaskOutcome.unwrap()` seems wrong, despite it being an example of 
        a user defined handle.
    - [x] @Fuyukai suggested `.wait_for_result()` which is a lot more
      pythonic in terms of plain-ol'-simple-semantics :surfer:
    - ~~maybe `.unwind()` is more descriptive? as in unwinding the outcome
      from the task would imply you both have to:~~
        - wait for the result or error from the underlying task
        - retrieve that outcome's value or raise it's error by unwrapping/boxing?
        - *stack unwinding* is normally synonymous with the end of
        a function call and the subsequent *popping of the stack* ..?
     - the nixed above ends up going down the rabbit hole of describing
        an imperative lang in terms of lower level fp abstractions which
        is not only confusing but a bit moot: we don't need to hear math
        about stuff we can't change in (python and) the semantics of
        a method..:joy:

- [ ] test the waters in `trio` core for interest in this idea
  - [x] had a decent discussion on it with the `anyio` author and njs but
    don't think they understood what i was proposing (hence this proto
    :joy:)
  - [ ] the main spot to start digging into how this might be implemented
    inside `trio`'s scheduler would be about here:
    https://github.com/python-trio/trio/blob/master/trio/_core/_run.py#L1601
    since this is where the nursery-global cancel scope's `CancelStatus`
    is activated on the currently spawning task.
    - obviously it's going to require quite a bit of reworking to try
      and offer the generator-style `@cancel_scope_manager` thing
      because we'll need to call `__enter__`/`__exit__` from two
      different `Runner` methods -> `.spawn_impl()` and `.task_exited()`
  - [ ] probably just eventually accept that no-one will comment on this
    PR (nor care?) and it'll get relegated to our `trionics` secret
    sauces pantry yet again :joy:

- [ ] figure out if it's possible to swap in **this** `open_nursery()`
  for all uses of `trio.open_nursery()` when our users enable debug mode
  :joy:
  - pretty sure we can do it at appropriate spots during (sub)actor boot?
    - root: https://github.com/goodboy/tractor/blob/master/tractor/_root.py#L135
    - child: https://github.com/goodboy/tractor/blob/master/tractor/_entry.py#L97

- [ ] review the `anyio` (WIP?) work around their version of an iface
  for this,
  - https://github.com/agronholm/anyio/pull/1098
    * https://github.com/agronholm/anyio/pull/1146
    * https://github.com/agronholm/anyio/pull/890
    * https://github.com/search?q=repo%3Aagronholm%2Fanyio+TaskHandle&type=pullrequests

- [ ] RECONDSIDER this whole task-oriented `tn` approach instead
     implementing the entire thing as just a `tn` of sub-`tn`s with
     each `tm.start[_soon]()` call simply opening another
     sub-`trio.Nursery` which would more or less provide the scope
     for "plugging-into" each single-`trio.Task`-under-lone-`tn`??

---

Maybe more to come?

### DIFF
--- a/tractor/trionics/_supervisor.py
+++ b/tractor/trionics/_supervisor.py
@@ -126,18 +126,23 @@ class ScopePerTaskNursery(Struct):
         # task = new_tasks.pop()
 
         n: Nursery = self._n
-        cs = CancelScope()
+
+        sm = self.scope_manager
+        # we do default behavior of a scope-per-nursery
+        # if the user did not provide a task manager.
+        if sm is None:
+            return n.start_soon(async_fn, *args, name=None)
+
+        # per_task_cs = CancelScope()
         new_task: Task | None = None
         to_return: tuple[Any] | None = None
 
-        sm = self.scope_manager
-        if sm is None:
-            mngr = nullcontext([cs])
-        else:
-            # NOTE: what do we enforce as a signature for the
-            # `@task_scope_manager` here?
-            mngr = sm(nursery=n)
-
+        # NOTE: what do we enforce as a signature for the
+        # `@task_scope_manager` here?
+        mngr = sm(
+            nursery=n,
+            # scope=per_task_cs,
+        )
         async def _start_wrapped_in_scope(
             task_status: TaskStatus[
                 tuple[CancelScope, Task]
@@ -148,48 +153,49 @@ class ScopePerTaskNursery(Struct):
             # TODO: this was working before?!
             # nonlocal to_return
 
-            with cs:
+            task = trio.lowlevel.current_task()
+            # self._scopes[per_task_cs] = task
 
-                task = trio.lowlevel.current_task()
-                self._scopes[cs] = task
+            # NOTE: we actually don't need this since the user can
+            # just to it themselves inside mngr!
+            # with per_task_cs:
 
-                # execute up to the first yield
-                try:
-                    to_return: tuple[Any] = next(mngr)
-                except StopIteration:
-                    raise RuntimeError("task manager didn't yield") from None
+            # execute up to the first yield
+            try:
+                to_return: tuple[Any] = next(mngr)
+            except StopIteration:
+                raise RuntimeError("task manager didn't yield") from None
 
-                # TODO: how do we support `.start()` style?
-                # - relay through whatever the
-                #   started task passes back via `.started()` ?
-                #   seems like that won't work with also returning
-                #   a "task handle"?
-                # - we were previously binding-out this `to_return` to
-                #   the parent's lexical scope, why isn't that working
-                #   now?
-                task_status.started(to_return)
+            # TODO: how do we support `.start()` style?
+            # - relay through whatever the
+            #   started task passes back via `.started()` ?
+            #   seems like that won't work with also returning
+            #   a "task handle"?
+            # - we were previously binding-out this `to_return` to
+            #   the parent's lexical scope, why isn't that working
+            #   now?
+            task_status.started(to_return)
 
-                # invoke underlying func now that cs is entered.
-                outcome = await acapture(async_fn, *args)
+            # invoke underlying func now that cs is entered.
+            outcome = await acapture(async_fn, *args)
 
-                # execute from the 1st yield to return and expect
-                # generator-mngr `@task_scope_manager` thinger to
-                # terminate!
-                try:
-                    mngr.send(outcome)
+            # execute from the 1st yield to return and expect
+            # generator-mngr `@task_scope_manager` thinger to
+            # terminate!
+            try:
+                mngr.send(outcome)
 
-                    # NOTE: this will instead send the underlying
-                    # `.value`? Not sure if that's better or not?
-                    # I would presume it's better to have a handle to
-                    # the `Outcome` entirely? This method sends *into*
-                    # the mngr this `Outcome.value`; seems like kinda
-                    # weird semantics for our purposes?
-                    # outcome.send(mngr)
 
-                except StopIteration:
-                    return
-                else:
-                    raise RuntimeError(f"{mngr} didn't stop!")
+                # I would presume it's better to have a handle to
+                # the `Outcome` entirely? This method sends *into*
+                # the mngr this `Outcome.value`; seems like kinda
+                # weird semantics for our purposes?
+                # outcome.send(mngr)
+
+            except StopIteration:
+                return
+            else:
+                raise RuntimeError(f"{mngr} didn't stop!")
 
         to_return = await n.start(_start_wrapped_in_scope)
         assert to_return is not None
@@ -198,7 +204,6 @@ class ScopePerTaskNursery(Struct):
         # mypy i guess..to like, relay the type of whatever the
         # generator yielded through? betcha that'll be un-grokable XD
         return to_return
-
 
 
 # TODO: you could wrap your output task handle in this?
@@ -214,6 +219,11 @@ class ScopePerTaskNursery(Struct):
 def add_task_handle_and_crash_handling(
     nursery: Nursery,
 
+    # TODO: is this the only way we can have a per-task scope
+    # allocated or can we allow the user to somehow do it if
+    # they want below?
+    # scope: CancelScope,
+
 ) -> Generator[None, list[Any]]:
 
     task_outcome = TaskOutcome()
@@ -222,8 +232,12 @@ def add_task_handle_and_crash_handling(
     task: Task = trio.lowlevel.current_task()
     print(f'Spawning task: {task.name}')
 
+    # yields back when task is terminated, cancelled, returns.
     try:
-        # yields back when task is terminated, cancelled, returns?
+        # XXX: wait, this isn't doing anything right since we'd have to
+        # manually activate this scope using something like:
+        # `task._activate_cancel_status(cs._cancel_status)` ??
+        # oh wait, but `.__enter__()` does all that already?
         with CancelScope() as cs:
 
             # the yielded value(s) here are what are returned to the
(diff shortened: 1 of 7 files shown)


tractor/trionics/_supervisor.py
```diff
@@ -0,0 +1,256 @@
...
+        await self._ready.wait()
+
+        out = self._outcome
+        if out is None:
+            raise ValueError(f'{out} is not an outcome!?')
+
+        return self.result
+
+
+class TaskHandle(Struct):
```
goodboy, line 87: TODO: show an second example where the `cancel_scope_manager`` generator wraps the trio internals listed as attrs into this compound type and then that is what's returned from the `.start_soon()` call?

tractor/trionics/_supervisor.py
```diff
@@ -0,0 +1,256 @@
...
+class ScopePerTaskNursery(Struct):
+    _n: Nursery
+    _scopes: dict[
+        Task,
+        tuple[CancelScope, Outcome]
+    ] = {}
+
+    scope_manager: ContextManager | None = None
+
+    async def start_soon(
```
goodboy, line 103: I don't really know what the best API would be for doing this with `.start()`?

1. you don't really need start since you can always implement it by getting the child task to `Event.set()`, or even further use this `MaybeOutcome` thing from above.
2. how do you get the `cancel_scope_manager` yielded output in that case, since `.start()` would normally return whatever the child passed to `status.started()`?
  - i.e. you can't do `started = await n.start(blah)` and expect `started` to be the `outcome, cs` stuff
3. maybe the child task could get access to it's own cancel scope "stuff" from somewhere and then choose to `status.started((outcome, cs))` it back to the caller?

tractor/trionics/_supervisor.py
```diff
@@ -0,0 +1,256 @@
...
+        ) as sn:
+            for _ in range(3):
+                outcome, cs = await sn.start_soon(trio.sleep_forever)
+
+            # extra task we want to engage in debugger post mortem.
+            err_outcome, *_ = await sn.start_soon(sleep_then_err)
+
+            val: str = 'yoyoyo'
+            val_outcome, cs = await sn.start_soon(sleep_then_return_val, val)
+            res = await val_outcome.unwrap()
```
goodboy, line 250: As already mentioned above in the comment on the method, probably a better name is warranted for this since you're actually first waiting for the outcome **then** unwrapping it's underlying value/result.

tractor/trionics/_supervisor.py
```diff
@@ -0,0 +1,256 @@
...
+            assert res == val
+            print(f'GOT EXPECTED TASK VALUE: {res}')
+
+            print('WAITING FOR CRASH..')
```
goodboy, line 254: scheduling `sleep_then_err()` results in an assertion error that is caught and handled in the `pdbp` REPL thanks to the `cancel_scope_manager` implementation which does this on std `Exception`s 😎 

Note, you could also catch `trio.Cancelled` or wtv else by changing the except line 🏄🏼

tractor/trionics/_supervisor.py
goodboy: Huh, i guess we could just offer some kinda support for hooking into the started value returned by the task? that way the `cancel_scope_manager` could decide what to send back to the `.start()` caller with the default being the value that the child `.started()` ?

goodboy: huh, likely you could implement that as a double yield generator then?

I dunno what would be best for ensuring that other then decorator params and then `async_fn` checking inside `.start()` against that i guess?

tractor/trionics/_supervisor.py
```diff
@@ -0,0 +1,256 @@
...
+        else:
+            mngr = sm(
+                nursery=n,
+                scope=cs,
+                maybe_outcome=maybe_outcome,
+            )
+
+        async def _start_wrapped_in_scope(
+            task_status: TaskStatus[
+                tuple[CancelScope, Task]
```
goodboy, line 139: wrong, this should be the type sig of whatever is yielded from the `self.scope_manager` obviously

tractor/trionics/_supervisor.py
```diff
@@ -0,0 +1,256 @@
...
+                    # `MaybeOutcome` type? Use `Outcome.send(mngr)`
+                    # right?
+                    maybe_outcome._set_outcome(outcome)
+
+        await n.start(_start_wrapped_in_scope)
+        assert to_return is not None
+
+        # TODO: better way to concat the values delivered by the user
+        # provided `.scope_manager` and the outcome?
+        return tuple([maybe_outcome] + to_return)
```
goodboy, line 177: yeah as mentioned in the TODO, don't love this concat-ing of the results from the `scope_manager()`, but i think this goes away if we instead implement our own generator as `@trio.cancel_scope_manager` protocol since then the absolute value `tuple` yielded can be returned directly?

tractor/trionics/_supervisor.py
```diff
@@ -0,0 +1,283 @@
...
+            task_status: TaskStatus[
+                tuple[CancelScope, Task]
+            ] = trio.TASK_STATUS_IGNORED,
+
+        ) -> None:
+
+            # TODO: this was working before?!
+            # nonlocal to_return
+
+            with cs:
```
goodboy, line 151: Ahh wait, right.. so we do need to activate this here or can it be done in the user's `@task_scope_manager` thing?

tractor/trionics/_supervisor.py
```diff
@@ -0,0 +1,283 @@
...
+        # NOTE: internals of a nursery don't let you know what
+        # the most recently spawned task is by order.. so we'd
+        # have to either change that or do set ops.
+        # pre_start_tasks: set[Task] = n._children.copy()
+        # new_tasks = n._children - pre_start_Tasks
+        # assert len(new_tasks) == 1
+        # task = new_tasks.pop()
+
+        n: Nursery = self._n
+        cs = CancelScope()
```
goodboy, line 129: Do we need this allocated here?
See comment(s) below.

tractor/trionics/_supervisor.py
```diff
@@ -0,0 +1,320 @@
...
+        await self._exited.wait()
+
+        out = self._outcome
+        if out is None:
+            raise ValueError(f'{out} is not an outcome!?')
+
+        return self.result
+
+
+class ScopePerTaskNursery(Struct):
```
goodboy, line 105: Heh, guess we don't need much of a special name now?

`PerTaskHookableNursery` 😂

tractor/trionics/_supervisor.py
```diff
@@ -0,0 +1,320 @@
...
+
+    scope_manager: Generator[Any, Outcome, None] | None = None
+
+    async def start_soon(
+        self,
+        async_fn,
+        *args,
+
+        name=None,
+        scope_manager: ContextManager | None = None,
```
goodboy, line 120: maybe `task_manager` is a better name?

tractor/trionics/_supervisor.py
```diff
@@ -0,0 +1,320 @@
...
+
+    ) -> tuple[CancelScope, Task]:
```
goodboy, line 122: needs to again be dynamically defined at runtime.. soo use the [`ParamSpec` stuff?](https://docs.python.org/3/library/typing.html#typing.ParamSpecArgs)

tractor/trionics/_supervisor.py
```diff
@@ -0,0 +1,317 @@
...
+                sleep_then_return_val,
+                val,
+            )
+            res = await val_outcome.wait_for_result()
+            assert res == val
+            print(f'{res} -> GOT EXPECTED TASK VALUE')
+
+            await trio.sleep(0.6)
+            print(
+                'Cancelling and waiting on {err_outcome.lowlevel_task} '
```
goodboy, line 312: lul, missing f-string

tractor/trionics/_supervisor.py
```diff
@@ -0,0 +1,317 @@
...
+        # TODO: use the fancy type-check-time type signature stuff from
+        # mypy i guess..to like, relay the type of whatever the
+        # generator yielded through? betcha that'll be un-grokable XD
+        return to_return
+
+
+# TODO: define a decorator to runtime type check that this a generator
+# with a single yield that also delivers a value (of some std type) from
+# the yield expression?
+# @trio.task_scope_manager
```
goodboy, line 206: maybe just `@trio.task_manager` ?

tractor/trionics/_supervisor.py
```diff
@@ -0,0 +1,317 @@
...
+def add_task_handle_and_crash_handling(
+    nursery: Nursery,
+
+    # TODO: is this the only way we can have a per-task scope
```
goodboy, line 210: no, we don't need the nursery to allocate any cancel scope per task interally!

need to remove this!

PR #373, opened 2025-03-21
Remote inceptions: improved `RemoteActorError` boxing of inter-actor exceptions

goodboy:
From the [introducing commit's msg](https://github.com/goodboy/tractor/pull/373/commits/c1e0328669f70cec92bb2ea784c184b9b4c45e33),

> Since adding more complex inter-peer (actor) testing scenarios, we
  definitely have an immediate need for `trio`'s style of "inceptions" but
  for nesting `RemoteActorError`s as they're relayed through multiple
  actor-IPC hops. So for example, a remote error relayed "through" some
  proxy actor to another ends up packing a `RemoteActorError` into another
  one such that there are 2 layers of RAEs with the first
  containing/boxing an original src actor error (type).

See that commit's very in depth explanation for more deats ;)

----

Also includes matching updates to the tests (which should mostly be
running green now :fingers_crossed:) and in particular extension of
the `tests/test_inter_peer_cancellation.py` suite.

### DIFF
--- a/examples/debugging/multi_daemon_subactors.py
+++ b/examples/debugging/multi_daemon_subactors.py
@@ -32,7 +32,7 @@ async def main():
             try:
                 await p1.run(name_error)
             except tractor.RemoteActorError as rae:
-                assert rae.type is NameError
+                assert rae.boxed_type is NameError
 
             async for i in stream:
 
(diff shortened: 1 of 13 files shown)


goodboy:
Yeah, no idea what's going on with actions and the `test_discovery` suite, but yet again it's all working local for me?

Issue #400, opened 2025-09-10
Distributed SC and remote cancellation, why `Cancelled` might not be enough?


goodboy:
From the original (shoddy) write-up, now reworked (and will be more) from
#387.

The whole point of this content is to attempt to explain the
different caveats that come with the idea of enforcing *distributed
structured concurrency* versus the traditional
`trio`-single-threaded-scheduler assumptions we all take for
granted..

Bp

Note that this is all a WIP and i'll be incrementally updating it as
my mind experiences *clear spells*.

---

> why this is a big deal for `tractor`'s RPC semantics, and in
> general the idea of "distributed SC"..

Remote exception "masking" is a bigger deal for us because we have to
worry about "remote error relay semantics" and maintaining SC rules
under various parallel, non-deterministic IPC msg arrival edge cases.

Particularly these (numbered) *premises*,

1. since there is no current way to override `trio.Cancelled`
   (ideally by plugging into `trio` core somehow) it forces us to use
   a combo of,

   - calling `Context._scope.cancel()` to signal and interrupt ongoing
     *user code* inside a `Portal.open_context()` or `@context` body,

   - expect any such `Context` embedded `trio.Task`s (children of
     `ctx.parent_task`) to handle `trio.Cancelled` despite its
     "underlying cause" not necessarily being a local error or
     `ctx.cancel()` call,

   - subsequently check `Context._scope.cancelled_caught` alongside
     other internal `Context` state before re-raising or "silently
     absorbing" any received remote error/`ContextCancelled` which
     was the "underlying cause" of the `Context`'s termination.
     * (eventually in the now released `trio>=0.31.0` we can customize
       the `Cancelled`'s `.reason: str` which may be a partial
       solution to the more general problem..)

   To both *perform* and/or *determine* the reason for
   a `Context`'s termination we need to handle any of a "side" being:

   - requested to be "gracefully cancelled" by `Context.cancel()`
     called in the peer task,

   - cancelled by the peer-actor's-task (the other "side") raising an
     exception,

   - transitively cancelled by some local error raised by a task
     in the same scope as the `Context.parent_task`; `Context._scope`
     is **not** `.cancel_called`.

   - cancelled due to a so called "out-of-band" (local or remote) request
     at a different "layer" of the runtime (see explainer below).

    Therefore, a task-cancellation (taskc) has **special meaning**
    when that taskc-requesting-scope **is specifically the**
    `Context._scope` and **when it is not**.

---

Understanding the existing runtime machinery..

[...]

Issue #388, opened 2025-07-17
Runtime-layer-aware cancel conditions, more `Cancelled` types and `checkpoint()` controls needed?


goodboy:
I've loosely commented on this in some spots in the code base prior but
never formally written anything and given the full justification for
the idea.

I drafted a bit of preface in the tail of #387 but here goes a more
rigorous justification..

It should be obvious that much of the *speculation* about (still
needed) lower level system features is **very** `linux` centric.

---

Some premises
----

- There are multiple runtime "layers" that exist for every
  `trio.Task` executing in a `tractor` program:

  - the task itself which is provided by `trio`-as-a-runtime.

  - the "actor" which is provided by `tractor` as the combo of an
    OS-process (and/or eventually shm-thread/sub-interpreter) + an embedded
    `trio.run()` (as well a booting a bunch of machinery to provide
    for all the functionality of that "actor").

  - the process (or shm-thread) hosting the python-interpreter
    depending on the spawning backend which can be
    customized/managed/controlled by various OS parameters (like
    namespaces and cgroups).

  - the bearing OS instance itself, particularly in the case where all
    default process settings are used from the prior bullet.

- Given the various layers having different "runtime scopes"
  there is a need to distinguish between them for the purposes of
  cancellation and error relay and detection, particularly to
  determine from a "higher-layer" (like a `trio.Task` defined in
  `tractor` user/app code) whether the "type" of a cancellation/error
  is due to some "lower-layer" (like say an OS SIGINT vs.
  a `Portal.cancel_actor()` request from some peer actor).

  * the main reason for desiring this, beyond just having explicit
    clarification of the source-of-origin of a remote/local
    cancellation/failure, is to enable *cancellation and failure
    handling* via a *supervisor* from different "layers" of the
    runtime (like from a [process][proc_super] vs. a [task manager][taskman]) as it pertains
    to (eventually) supporting these (sometimes overlapping) general
    properties in `tractor`-based distributed programs,
    - [availability](https://en.wikipedia.org/wiki/High_availability)
    - [adaptability](https://en.wikipedia.org/wiki/Adaptive_system)
    - [resiliency](https://en.wikipedia.org/wiki/Resilience_engineering)
    - [reliability](https://en.wikipedia.org/wiki/Reliability_engineering)
    - [survivability](https://en.wikipedia.org/wiki/Survivability)

  * IMO the easiest way to handle such cancels/faults in
    a distributed SC program (as we'd like to propose are easy to
[...]

PR #403, opened 2025-09-12
idea: an `ActorCancelled(ContextCancelled)` exc type for better OoB cancelled handling

goodboy:
Per recent testing effort, issue-docs drafting and generally ranting in the `trio` matrix room 🫣 here is the recent content driving this idea,

- #399 starts dedicated testing around such scenarios
- #400 attempts to justify the need for more then just `trio.Cancelled`
- #391 demos why `trio.Cancelled` and so called "exception masking" can be problematic in a distributed system where error/cancel relay is important for distributed-sys feats.

---

### New supporting feats in `trio` we should try here

With [release `0.31.0`](https://github.com/python-trio/trio/releases/tag/v0.31.0) we get a new `Cancelled.reason: str` which can be set at the `.cancel(reason)` call,

- introducing patch, https://github.com/python-trio/trio/issues/3232
- docs on [`CancelScope.cancel()`](https://trio.readthedocs.io/en/stable/reference-core.html#trio.CancelScope.cancel)
- low level impl details,
  * internal [`CancelReason` struct](https://github.com/python-trio/trio/blob/main/src/trio/_core/_run.py#L323)
  * which is now a [private instance var](https://github.com/python-trio/trio/blob/main/src/trio/_core/_run.py#L588) on all scopes.
  * how and where it's set in [`._cancel()](https://github.com/python-trio/trio/blob/main/src/trio/_core/_run.py#L918)
  * example low-level-internal usage in case we wanted to overload the type in [`._child_finished()`](https://github.com/python-trio/trio/blob/main/src/trio/_core/_run.py#L1273)

### DIFF
--- a/examples/debugging/multi_nested_subactors_error_up_through_nurseries.py
+++ b/examples/debugging/multi_nested_subactors_error_up_through_nurseries.py
@@ -21,12 +21,12 @@ async def breakpoint_forever():
 async def spawn_until(depth=0):
     """"A nested nursery that triggers another ``NameError``.
     """
-    async with tractor.open_nursery() as n:
+    async with tractor.open_nursery() as an:
         if depth < 1:
 
-            await n.run_in_actor(breakpoint_forever)
+            await an.run_in_actor(breakpoint_forever)
 
-            p = await n.run_in_actor(
+            p = await an.run_in_actor(
                 name_error,
                 name='name_error'
             )
@@ -38,7 +38,7 @@ async def spawn_until(depth=0):
             # recusrive call to spawn another process branching layer of
             # the tree
             depth -= 1
-            await n.run_in_actor(
+            await an.run_in_actor(
                 spawn_until,
                 depth=depth,
                 name=f'spawn_until_{depth}',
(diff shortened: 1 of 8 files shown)


goodboy:
> archival note authored by `opencode`

Closing this draft in favor of #515, which now tracks the unresolved
`ActorCancelled` relay and nursery-exit contract independently from this
prototype.

The durable parts of this work are preserved there:

- actor-runtime cancellation needs a signal distinct from local
  `trio.Cancelled` and task-level `ContextCancelled`;
- graceful actor cancellation must be distinguished from error-triggered
  one-cancels-all teardown;
- the triggering error and one cancellation outcome per affected sibling
  need a defined grouping policy;
- canceller identity and relay metadata must survive process boundaries.

This branch should not be rebased wholesale. Its supervision rewrite left
real correctness problems unresolved: parent-generated cancellation could
race with and replace a real `RemoteActorError`, cancellation outcomes did
not reliably escape `trio.run()`, and the infinite-stream cancellation path
could hang. The branch also retains commented-out injection paths, incomplete
error collection, and diagnostic test scaffolding.

#388 remains the broader layered-cancellation design, #400 records the
distributed cancellation rationale, and #515 is the bounded follow-up for
specifying and testing the actor-level contract. This PR remains useful as
implementation prior art and as a record of approaches that should not be
repeated unchanged.

PR #414, opened 2026-02-20
Hide private fields in `Struct.pformat()` output

goodboy:
Skip fields starting with `_` in pretty-printed struct output to
avoid cluttering console with internal/private state (and/or
accessing private properties which raise errors.

Impl deats,
- add `if k[0] == '_': continue` check to skip private fields
- change nested `if isinstance(v, Struct)` to `elif` since we now
  have early-continue for private fields
- mv `else:` comment to clarify it handles top-level fields
- fix indentation of `yield` statement to only output non-private,
  non-nested fields

---

### Add `tests/msg/` sub-pkg, audit `pformat()` filtering

Reorganize existing msg-related test suites under a new
`tests/msg/` subdir (matching `tests/devx/` and `tests/ipc/`
convention) and add unit tests for the `_`-prefixed field
filtering in `pformat()`.

Deats,
- `git mv` `test_ext_types_msgspec` and `test_pldrx_limiting`
  into `tests/msg/`.
- add `__init__.py` + `conftest.py` for the new test sub-pkg.
- add `test_pretty_struct.py` with 8 unit tests:
  - parametrized field visibility (public shown,
    `_`-private hidden, mixed)
  - direct `iter_struct_ppfmt_lines()` assertion
  - nested struct recursion filtering
  - empty struct edge case
  - real `MsgDec` via `mk_dec()` hiding `_dec`
  - `repr()` integration via `Struct.__repr__()`

---

### Pre-merge todo

- [x] unit test verifying **only** non-private fields are
  rendered in `repr()` output.
- [x] reorg existing `tractor.msg` test files under
  `tests/msg/` subdir.
- [x] CI green on all new + moved test suites.

---

### Follow up

moved to #425.

### DIFF
--- a/tests/msg/test_pretty_struct.py
+++ b/tests/msg/test_pretty_struct.py
@@ -3,10 +3,6 @@ Unit tests for `tractor.msg.pretty_struct`
 private-field filtering in `pformat()`.
 
 '''
-from typing import (
-    Any,
-)
-
 import pytest
 
 from tractor.msg.pretty_struct import (


goodboy:
Pretty sure landing #422 first means CI will be green.. so maybe i'll rebase 🤷🏼

PR #446, opened 2026-04-21
A subinterpreter-in-thread spawning backend

goodboy:
### Motivation

This PR is **Phase B.2** of the multi-backend spawn rework tracked
under #379: it lands the first member of the `subint` family — a
single-process spawn backend that runs each sub-actor inside a
CPython PEP 734 sub-interpreter driven by a dedicated OS thread,
while reusing tractor's existing IPC handshake (UDS/TCP) for the
parent↔child channel. The result is multiple `Actor`s in **one OS
process** with state isolation and faster startup than
`mp_spawn`/`mp_forkserver`, but **without** the per-interp-GIL
parallelism that's still gated upstream on `msgspec` PEP 684 support
(jcrist/msgspec#563 and more specifically jcrist/msgspec#1026).

Why the dedicated-thread machinery? PEP 734's public
`concurrent.interpreters` only exposes `'isolated'` config, which
refuses to import any C extension missing the
`Py_mod_multiple_interpreters` slot — and that's currently `msgspec`,
which tractor uses pervasively in the IPC layer. So we drop to the
**private** `_interpreters` C module in `'legacy'` config: shared
main GIL, but the `sys.modules`/`__main__`/globals isolation we
actually want for the actor model. We also feature-gate to **3.14+**
via the public-module presence check: the private API exists on 3.13,
but tractor's multi-trio-task usage of `_interpreters.exec()` wedges
silently there and only works cleanly on 3.14.

This PR also documents two known hang classes specific to in-thread
subints (a class-A SIGINT-starvation hang and an orphaned-channel
park) and ships the `skipon_spawn_backend()` pytest marker +
`tractor.devx._debug_hangs` module that fell out of triaging them. It
serves as the foundation for the stacked subint-family PRs (#447
forkserver, future fork variants).

---

### Src of research

- PEP 734 — `concurrent.interpreters` public API
  * https://peps.python.org/pep-0734/
- PEP 684 — per-interpreter GIL / `Py_mod_multiple_interpreters`
  * https://peps.python.org/pep-0684/
- CPython public wrapper (`Lib/concurrent/interpreters/`) *
  https://github.com/python/cpython/tree/main/Lib/concurrent/interpreters
- CPython private C ext (`Modules/_interpretersmodule.c`) *
  https://github.com/python/cpython/blob/main/Modules/_interpretersmodule.c
- `msgspec` PEP 684 upstream tracker — jcrist/msgspec#563
- anyio's `to_interpreter._Worker` for a `set___main___attrs()`
[...]

Issue #379, opened 2025-07-07
Trying out sub-interpreters (subints), maybe `fork()` can be hacked now?'


goodboy:
As per the new pub API dropping in 3.14,

- https://docs.python.org/3.14/library/concurrent.interpreters.html
  - https://docs.python.org/3.14/library/concurrent.interpreters.html#basic-usage
  - https://peps.python.org/pep-0734/

- https://docs.python.org/3.14/library/concurrent.futures.html#interpreterpoolexecutor
  > Each worker’s interpreter is isolated from all the other
  > interpreters. “Isolated” means each interpreter has its own
  > runtime state and operates completely independently. For example,
  > if you redirect sys.stdout in one interpreter, it will not be
  > automatically redirected to any other interpreter. If you import
  > a module in one interpreter, it is not automatically imported in
  > any other. You would need to import the module separately in
  > interpreter where you need it. In fact, each module imported in
  > an interpreter is a completely separate object from the same
  > module in a different interpreter, including sys, builtins, and
  > even __main__.

The latter explanation more-or-less implies our existing runtime
should work OTB i think?

It'd sure be cool to get (close to) multi-threading like spawn
latency as well as see if we can work around the known limitations
with forking and `trio` due to runtime duplication and state
conflicts (see discussion below).

---

Existing `trio` issues,
----------------------
- directly regarding `fork()`-ing,
  - https://github.com/python-trio/trio/issues/1614
  - https://github.com/python-trio/trio/issues/608
  - https://github.com/python-trio/trio/issues/2764
  - https://github.com/python-trio/trio/issues/1607

Pre-existing `subint`  integrations (in the 3rd-party ecosys)
-----------------------------------

- `trio-parallel` patch adding support,
  - https://github.com/richardsheridan/trio-parallel/compare/main...sint

- `anyio`'s support since 3.13,
  - docs: https://anyio.readthedocs.io/en/stable/subinterpreters.html
  - question about async task in subint: https://github.com/agronholm/anyio/discussions/904
    - user's current approach to run web-server-threads:
      https://github.com/coreproject-moe/CoreProject/blob/d8eb3a7e212c17ec3534f7f9486f6040cb7a6f71/tracker/src/coreproject_tracker/cli.py#L28-L49

---

Our own thoughts, ideas for `fork()`-workaround/hacks..
----------------------

I had previously schemed (that prior to going full gil-ess python)
we'd need to write our own fork-server (since the stdlib's has many
[...]

PR #447, opened 2026-04-22
Working toward a "subinterpreter-forkserver" spawning backend

### LINKED ISSUE #456: Spawn-time boot-death (`rc=2`) under rapid same-name spawn against a registrar
Pertains to recent `main_thread_forkserver` fixes from the WIP #447
work, namely commit 3b0724e.

---

## Symptom

Spawning N (≥4) sub-actors with the **same name** in tight
succession against a daemon registrar surfaces as
`ActorFailure: Sub-actor (...) died during boot (rc=2)
before completing parent-handshake`.

```
tests/discovery/test_multi_program.py
  ::test_dup_name_cancel_cascade_escalates_to_hard_kill[n_dups=4]
```

```
tractor._exceptions.ActorFailure:
  Sub-actor ('doggy', '<uuid>') died during boot (rc=2)
  before completing parent-handshake.
    proc: <_ForkedProc pid=<n> returncode=None>
```

The `proc` repr shows `returncode=None` because the repr is
captured before `proc.wait()` returns; the actual
`os.WEXITSTATUS == 2` is reported via `result['died']` in the
race-helper.

## When it surfaces

- N=2 (`n_dups=2`): **always passes**.
- N=4 (`n_dups=4`): **consistent fail** under both `tpt-proto=tcp`
  and `tpt-proto=uds`, MTF backend.
- N=8 (`n_dups=8`): **passes** (counter-intuitive — see "racing
  windows").
- Non-MTF backends: not yet exercised systematically.

## What previously masked it

Pre the spawn-time `wait_for_peer_or_proc_death` race-helper
(in `tractor.spawn._spawn`), the parent's `start_actor` flow
ended with a bare:

```python
event, chan = await ipc_server.wait_for_peer(uid)
```

That awaits an unsignalled `trio.Event` on `_peer_connected[uid]`.
If the sub-actor process **dies during boot** (before its
runtime executes the parent-callback handshake that sets the
event), the wait parks forever. The dead proc becomes a zombie
because no one ever calls `proc.wait()` to reap it.

In test contexts the failure presented as a hang or a much
later `trio.TooSlowError` from an outer `fail_after`. In
production it'd present as a parent that never makes progress
past `start_actor`. The death itself was silently masked.

## What surfaces it now

`tractor.spawn._spawn.wait_for_peer_or_proc_death` (used by
`_main_thread_forkserver_proc`) races the handshake-wait
against `proc.wait()`. The race-helper raises `ActorFailure`
on death-first instead of parking, exposing the rc=2.

## Hypothesis: registrar-side same-name contention

The test spawns N actors with name `doggy` sequentially:

```python
for i in range(n_dups):
    p: Portal = await an.start_actor('doggy')
    portals.append(p)
```

Each spawned doggy:

1. Forks via the forkserver.
2. Boots its runtime in `_actor_child_main`.
3. Connects back to the parent for handshake.
4. Connects to the daemon registrar to call `register_actor`.
5. Enters its RPC msg-loop.

Step (4) is where the same-name contention lives. The
registrar's `register_actor` (in
`tractor.discovery._registry`) accepts duplicate names
(stores `(name, uuid) -> addr`), but its internal bookkeeping
may have a non-trivial check (e.g. `wait_for_actor` resolution,
`_addrs2aids` map updates) that errors out under specific
ordering between the existing entry and the incoming one.

`rc=2 == os.WEXITSTATUS == 2` corresponds to `sys.exit(2)`
in the doggy process — typically reached via an unhandled
exception that's translated to exit code 2 by Python's top-
level (e.g. `argparse` errors use 2; `SystemExit(2)` etc.).
So the doggy is hitting an explicit exit path during
`register_actor` or just-after.

The non-monotonic shape (N=2 OK, N=4 BAD, N=8 OK) suggests a
specific timing window — likely "the 3rd register-RPC arrives
while the 1st-or-2nd is in some intermediate state". With
N=8, the additional procs widen the registration spread
enough that no two land in the conflicting window.

## Where to dig next

- Add per-actor logging in `_actor_child_main` and
  `register_actor` to surface the actual exception that
  triggers the rc=2 exit. Currently the doggy dies before
  the parent ever sees its stderr (forkserver doesn't
  marshal child stdio back).
- Race-test the registrar's `register_actor` /
  `unregister_actor` /  `wait_for_actor` against same-name
  concurrent calls in isolation (no spawn).
- Consider whether `register_actor` should be idempotent
  under same-name re-register or should explicitly reject
  same-name (and ideally with a clear `RemoteActorError`,
  not `sys.exit(2)`).

## Test-suite handling

Currently:

- `tests/discovery/test_multi_program.py
  ::test_dup_name_cancel_cascade_escalates_to_hard_kill[n_dups=4]`
  is `pytest.mark.xfail(strict=False, reason=...)` to keep
  the suite green while this issue is investigated.
- `n_dups=2` and `n_dups=8` continue to validate the
  cancel-cascade hard-kill escalation.

Once the underlying race is understood + fixed, drop the
xfail.

## Related work

- The cancel-cascade fix that introduced this regression
  test:
  `tractor/_exceptions.py:ActorTooSlowError`,
  `tractor/runtime/_supervise.py:_try_cancel_then_kill`,
  `tractor/runtime/_portal.py:Portal.cancel_actor(
   raise_on_timeout=...)`.
- The spawn-time death-detection that exposed this:
  `tractor/spawn/_spawn.py:wait_for_peer_or_proc_death`,
  used by `tractor/spawn/_main_thread_forkserver.py`.

goodboy:
## Add `subint_forkserver` spawn backend (#379 follow-on)

(i can't believe how fast we vibed this 😂 )

### Motivation

Stacked on top of [PR #446 / `subint_fork_backend`][pr-446]: that
branch laid the `_subint_fork.py` stub and the post-fork-CPython
exploration showing that `os.fork()` from a non-main subinterpreter
aborts the child at the CPython layer. This branch operationalizes
the workaround sketched there — fork from a regular
`threading.Thread` attached to the *main* interpreter (one that has
never entered a subint) — and wires it through tractor as a
first-class spawn backend.

The implementation looks tiny (one new module) but the supporting
work was the bulk of the patch series: multiple cancel-cascade hangs
surfaced under fork-based teardown that didn't exist for any
exec-based backend (`trio`, `mp_*`). The shared-memory image
inherited across `os.fork()` makes parent↔child socket-EOF delivery
racy, which exposed a latent `process_messages`-shielded-loop
deadlock; once cracked, that fix benefits every backend. A separate
`pytest --capture=fd` × fork-child interaction was traced to a
`test_nested_multierrors` cancel-cascade hang gated on the capture
mode — tracked at #449 and worked-around by defaulting the suite to
`--capture=sys`.

Late in the series the backend was split into two clearly-named
variants: **variant 1** — `main_thread_forkserver` — is the
working backend that ships today (forks from a regular main-interp
worker thread, child runs trio on its own main interp; NO
subinterpreter anywhere); **variant 2** — `subint_forkserver` —
is reserved as a placeholder for the future subint-isolated child
runtime, gated on jcrist/msgspec#1026 (PEP 684 isolated-mode
support). Today the `'subint_forkserver'` spawn-method key
dispatches to a `NotImplementedError` stub that points operators at
the variant-1 key. The "subint" prefix on both modules is
family-naming — they live alongside `_subint.py` /
`_subint_fork.py` from the broader #379 series.

The two `threading.Thread` primitives in `_main_thread_forkserver`
are deliberately heavy-handed (full ad-hoc threads, not
`trio.to_thread.run_sync`) to side-step legacy-config-subint GIL
starvation; once `msgspec` lands PEP 684 support and we can use
isolated subints, that constraint relaxes — auditable revisit
tracked at #450. Also bundled: a `tractor-reap` zombie-subactor
cleanup CLI + `_testing._reap` shared impl + session-scoped autouse
fixture, so a mid-teardown timeout no longer leaves orphan
subactors competing for ports across test sessions; a follow-up
commit extends `tractor-reap` with a `--shm` mode that sweeps
orphaned `/dev/shm/*` segments owned by the current uid that no
live process is mapping or holding open.

---

### Src of research

The following provide info on why/how this impl makes sense,

- [goodboy/tractor#379](https://github.com/goodboy/tractor/issues/379)
  * motivating issue: "Trying out sub-interpreters (subints), maybe
    `fork()` can be hacked now?". The "Our own thoughts" section
    sketches the worker-thread fork pattern that this backend
    implements.
- [PR #446 — `subint_fork_backend`][pr-446]
  * the immediate base branch — establishes the post-fork-CPython
    block and the `_subint_fork.py` stub returning
    `NotImplementedError`.
- <https://peps.python.org/pep-0684/>
  * per-interpreter GIL via `Py_mod_multiple_interpreters`. Until
    `msgspec` adopts the slot we're stuck on legacy-config subints,
    which forces our heavier thread design (see #450).
- [jcrist/msgspec#1026](https://github.com/jcrist/msgspec/issues/1026)
  * upstream `msgspec` PEP 684 isolated-mode tracker; gates
    variant-2.
- `ai/conc-anal/subint_fork_from_main_thread_smoketest.py`
  * standalone CPython-level reproducer covering the four scenarios
    (`control_subint_thread_fork`, `main_thread_fork`,
    `worker_thread_fork`, `full_architecture`) — pre-tractor proof
    that the workaround is sound.

**Module-level design docs** (read these top-of-file docstrings
for the per-backend architectural justifications, fork-semantics
analysis, and migration plans — much richer than the
high-level summary in this PR description):

- [`tractor/spawn/_main_thread_forkserver.py`][_main_thread_forkserver-py]
  * variant-1 backend (working today). Sections: Background,
    Design rationale (why a forkserver + why in-process),
    What survives the fork? — POSIX semantics, FYI: how this
    dodges the `trio.run()` × `fork()` hazards, Implementation
    status, Still-open work, TODO gated on msgspec PEP 684.
- [`tractor/spawn/_subint_forkserver.py`][_subint_forkserver-py]
  * variant-2 placeholder. Sections: Future arch — what subints
    would buy us (3 wins: cheaper forks, true parallelism,
    multi-actor-per-process), what lives here today
    (`run_subint_in_worker_thread`), what will live here when
    variant 2 ships.
- [`tractor/spawn/_subint.py`][_subint-py]
  * in-thread `subint` backend (parent of this stack, PR #446).
    Why we use the *private* `_interpreters` C module instead of
    `concurrent.interpreters`'s public `'isolated'` API; py3.14+
    feature gate rationale; msgspec PEP 684 migration path.
- [`tractor/spawn/_subint_fork.py`][_subint_fork-py]
  * fork-from-non-main-subint RFC stub (blocked upstream).
    Pointers to the CPython-source-line analysis in
    `subint_fork_blocked_by_cpython_post_fork_issue.md`.

---

### Summary of changes

By chronological commit,

- ([82332fbc][82332fbc]) Lift the validated fork primitives into
  `tractor.spawn._subint_forkserver`: `fork_from_worker_thread()` +
  `run_subint_in_worker_thread()` as the two re-usable building
  blocks.

- ([25e400d5][25e400d5]) Add trio-parent integration tests covering
  tier-1 (primitives driven from inside `trio.run()`) and tier-2
  (full backend wired through `open_root_actor` + `open_nursery`).

- ([cf2e71d8][cf2e71d8]) Document the PEP 684 audit-plan under
  `ai/conc-anal/subint_forkserver_thread_constraints_on_pep684_issue.md`
  — the upstream-gated cleanup work tracked at #450.

- ([26914fde][26914fde]) Wire `'subint_forkserver'` as a first-class
  `SpawnMethodKey` and `_methods` registry entry; the
  `try_set_start_method` case re-uses the subint-family py3.14+
  gate.

- ([63ab7c98][63ab7c98]) ([7804a9fe][7804a9fe]) Reset post-fork
  `_state` in the forkserver child via a new pure
  `get_runtime_vars(clear_values=True)` + sibling
  `set_runtime_vars()` API; without the reset the child inherits
  the parent's `_is_root=True` and trips `Actor._from_parent()` on
  the `SpawnSpec` handshake.

- ([76605d56][76605d56]) ([dcd5c1ff][dcd5c1ff])
  ([a72deef7][a72deef7]) Add a DRAFT orphan-SIGINT test scaffold +
  `child_sigint` modes; refine the diagnosis — the hang is NOT a
  missing handler, trio's loop stays wedged in `epoll_wait` despite
  delivery. Full trace + fix dirs in
  `ai/conc-anal/subint_forkserver_orphan_sigint_hang_issue.md`.

- ([f5f37b69][f5f37b69]) ([5e85f184][5e85f184])
  ([8bcbe730][8bcbe730]) ([e31eb8d7][e31eb8d7]) Shorten timeouts
  in forkserver suites, drop dead f-string prefixes, enable
  `debug_mode` for the forkserver path (added to a new
  `_DEBUG_COMPATIBLE_BACKENDS` list in `_root`), and label the
  forkserver child in log attribution.

- ([1e357dcf][1e357dcf]) Mv `test_subint_cancellation.py` into the
  new `tests/spawn/` subpkg alongside the forkserver test module.

- ([d093c319][d093c319]) ([70d58c4b][70d58c4b]) Teach the
  `/run-tests` skill a zombie-actor post-run check + a SIGINT-first
  graceful cleanup ladder. Per the SC-discipline rule: graceful
  cancel before SIGKILL.

- ([e3f4f5a3][e3f4f5a3]) ([1af21210][1af21210]) Add the
  test-cancellation leak doc
  (`subint_forkserver_test_cancellation_leak_issue.md`) and wire
  `reg_addr` fixture through the leaky cancel tests so each run
  gets a unique registrar address.

- ([35da8089][35da8089]) ([9993db01][9993db01]) Refine the
  nested-cancel hang diagnosis; add a post-fork FD scrub in the
  fork-child prelude as the current workaround.

- ([c20b05e1][c20b05e1]) Use `pidfd` for cancellable
  `_ForkedProc.wait()` — replaces a blocking `os.waitpid()` with a
  `trio`-cancellable poll.

- ([8ac3dfeb][8ac3dfeb]) Break the parent-channel shield in
  `Actor.cancel()` teardown via a captured `_parent_chan_cs`.
  Without this, the shielded `process_messages` loop parks on EOF
  that only arrives AFTER the parent tears down — under fork
  backends the parent is itself blocked on this child's exit.
  Mutual-wait deadlock; explicit cancel makes teardown deterministic
  regardless of backend.

- ([506617c6][506617c6]) ([ab86f761][ab86f761])
  ([458a35cf][458a35cf]) ([7cd47ef7][7cd47ef7]) Skip-mark + narrow
  the cancel hang; surface silent failures in the forkserver child;
  doc ruled-out fixes + the capture-pipe aside.

- ([76d12060][76d12060]) Claude-perms tweak so `/commit-msg` outputs
  can be written.

- ([4106ba73][4106ba73]) ([eceed29d][eceed29d])
  ([4c133ab5][4c133ab5]) Pin the forkserver hang to `pytest
  --capture=fd` (#449); codify the capture-pipe-hang lesson in
  skills; default `pytest` to `--capture=sys` in `pyproject.toml`
  with the trade-off rationale inlined.

- ([e312a68d][e312a68d]) ([4d055543][4d055543]) Bound the peer-clear
  wait in `async_main`'s `finally` (3s `move_on_after`) and narrow
  the forkserver hang to the `async_main` outer tn — load-bearing
  for backend-agnostic teardown determinism.

- ([d6e70e9d][d6e70e9d]) Import-or-skip `.devx.` tests requiring
  `greenback` — keeps the suite collectable without the optional
  dep.

- ([b350aa09][b350aa09]) Wire `reg_addr` through infected-asyncio
  tests for parallel-run isolation.

- ([2ca0f41e][2ca0f41e]) ([44bdb169][44bdb169]) Skip
  `test_loglevel_propagated_to_subactor` on the forkserver backend
  too; tighten the orphan-SIGINT xfail to `strict=True`.

- ([eae478f3][eae478f3]) ([6d76b604][6d76b604]) Add
  `tractor._testing._reap` (SC-polite SIGINT-first reap, descendant
  + orphan modes) + the session-scoped autouse
  `_reap_orphaned_subactors` fixture; add the `tractor-reap` CLI
  (`scripts/tractor-reap`) wrapping the same impl.

- ([c99d475d][c99d475d]) ([aa3e2309][aa3e2309]) Fix
  `mp.SharedMemory` under fork-without-exec —
  `tractor.ipc._mp_bs.disable_mantracker(force_disable=True)` is
  now the default (belt+suspenders: no-op `ManTracker` monkey-patch
  + `track=False`); `_shm.open_shm_list` always wires the `unlink`
  lifetime callback (was 3.12-and-below only); document the
  incompat in
  `ai/conc-anal/subint_forkserver_mp_shared_memory_issue.md`.

- ([4f12d69b][4f12d69b]) Extend `tractor-reap` with `--shm` (and
  `--shm-only`) modes that sweep orphaned `/dev/shm/<key>` segments
  owned by the current uid with no live process mapping or holding
  them open. Match-criteria via `psutil.Process.memory_maps()` +
  `.open_files()` — kernel-canonical, no reliance on
  tractor-specific shm-key naming, so unrelated apps' segments are
  always preserved. Adds `psutil>=7.0.0` to the `testing` dep
  group.

- ([65fcfbf2][65fcfbf2]) ([9b05f659][9b05f659])
  ([66f1941f][66f1941f]) Bump `test_stale_entry_is_deleted` timeout
  to 30s; wire `test_dynamic_pub_sub` to standard fixtures; wire
  `reg_addr` into `test_context_stream_semantics`.

- ([54561959][54561959]) Surface subint bootstrap excs in
  `_subint.subint_proc()` (`try/except BaseException` +
  `log.exception(...)` around `_interpreters.exec()`); also log
  `_interpreters.is_running(interp_id)` on hard-kill timeout to
  disambiguate "thread leaked, subint already done" from "thread
  alive bc subint is wedged". `?TODO` notes the anyio-borrow path
  for re-raising bootstrap excs in the parent task and migrating to
  `_interpreters.set___main___attrs()` for non-literal SpawnSpec
  args.

- ([3ab99d55][3ab99d55]) ([4b5176e2][4b5176e2]) Major
  module-docstring expansion for `_subint_forkserver`: design
  rationale (in-process forkserver vs. `mp.forkserver`'s sidecar;
  why a forkserver at all vs. forking from a trio task), POSIX/trio
  fork mechanics, what survives the fork boundary, and the
  future-subint payoffs (cheaper forks, true parallelism via
  per-interp GIL, multi-actor-per-process). Bump the gated msgspec
  link from `#563` → `#1026`.

- ([99dade0f][99dade0f]) Extract the truly-generic
  main-interp-worker-thread fork primitives
  (`fork_from_worker_thread`, `_close_inherited_fds`, `_ForkedProc`,
  `wait_child`, `_format_child_exit`) into a sibling
  `_main_thread_forkserver.py` module — the primitive layer is now
  honestly named (none of these helpers touch a subint). Re-exports
  preserved.

- ([57dae0e4][57dae0e4]) Split the backend into variant 1 + variant
  2 modules. Variant 1 (`main_thread_forkserver`) becomes the
  canonical working impl: new `SpawnMethodKey` literal, `_methods`
  dispatch entry, `Actor._from_parent()` match-arm,
  `main_thread_forkserver_proc()` spawn-coro stamping its own
  `SpawnSpec` / log lines. Variant 2 (`subint_forkserver`) shrinks
  to a placeholder describing the future subint-isolated child
  runtime gated on jcrist/msgspec#1026; legacy `'subint_forkserver'`
  key still aliases to variant-1 here (flipped to
  `NotImplementedError` in the next commit).

- ([5e83881f][5e83881f]) Reduce `_subint_forkserver.py` to its
  variant-2 placeholder shape: add `subint_forkserver_proc` async
  stub raising `NotImplementedError` with a redirect msg pointing
  at `main_thread_forkserver` + jcrist/msgspec#1026 + #379. Flip
  the `_methods` registry to dispatch the stub directly so
  `--spawn-backend=subint_forkserver` errors cleanly. Drop dead
  module-scope (`ChildSigintMode`, `_DEFAULT_CHILD_SIGINT`, unused
  imports).

- ([9f0709ee][9f0709ee]) Rename
  `tests/spawn/test_subint_forkserver.py` →
  `test_main_thread_forkserver.py`; migrate test/smoketest imports
  to `tractor.spawn._main_thread_forkserver`; orphan-harness
  subprocess argv flipped to `'main_thread_forkserver'`. Drop the
  variant-2 module's backward-compat re-exports of fork primitives.

- ([205382a3][205382a3]) Sweep `subint_forkserver` →
  `main_thread_forkserver` in remaining string-match refs:
  `_DEBUG_COMPATIBLE_BACKENDS`,
  `test_loglevel_propagated_to_subactor`'s capfd-skip,
  `test_sigint_closes_lifetime_stack`'s xfail, comment/docstring
  refs across `_runtime`, `_state`, `_testing.pytest`, `_subint`,
  `pyproject.toml`, `test_cancellation`, `test_registrar`. Drop the
  `test_shm.py` "broken on `main_thread_forkserver`" skip-mark —
  `_mp_bs` + `_shm` fixes make those tests pass.

- ([cbdf1eb6][cbdf1eb6]) Add
  `test_subint_forkserver_key_errors_cleanly` regression guard
  pinning the variant-2 reservation contract: the
  `'subint_forkserver'` key MUST raise `NotImplementedError` (not
  silently dispatch to variant-1), and the error msg must surface
  both the working-backend pointer (`main_thread_forkserver`) +
  the upstream blocker (`msgspec#1026`).

- ([7c5dd4d0][7c5dd4d0]) Fix `_testing.addr.get_rando_addr`
  cross-process collisions: the `_rando_port: str =
  random.randint(...)` default-arg expression was evaluated ONCE at
  module-import — making it a per-process singleton. Two parallel
  pytest sessions had a 1/9000 birthday-pair chance of cascade-
  failing every `reg_addr`-using test. Switch to per-call
  `random.randint()` salted with `os.getpid()`; drop the bogus
  `: str` annotation.

---

### Future follow up

- [ ] Resolve `test_nested_multierrors` cancel-cascade hang under
  `--capture=fd` (#449). The `--capture=sys` default is a
  workaround; the underlying pytest-capture-machinery ↔ fork-child
  stdio interaction is not yet root-caused. See
  `ai/conc-anal/subint_forkserver_test_cancellation_leak_issue.md`.

- [ ] Wire the variant-2 `subint_forkserver_proc()` impl once
  `msgspec` ships PEP 684 isolated-mode support
  (jcrist/msgspec#1026 → tracked at #450). Today it's a
  `NotImplementedError` stub; the unblock lets the child runtime
  live in an isolated subint while the parent's `trio.run()` keeps
  running on main.

- [ ] Audit `main_thread_forkserver` thread-constraint cleanup once
  `msgspec` ships PEP 684 support (#450). Both primitives currently
  allocate dedicated `threading.Thread` instances rather than using
  `trio.to_thread.run_sync`; the `TODO — cleanup gated on msgspec
  PEP 684 support` block in the module docstring catalogs the three
  entangled root causes that block the cleanup today.

- [ ] Surface subint bootstrap exceptions to the parent task via a
  `nonlocal err` slot. `_subint.subint_proc()` currently logs them
  via `log.exception()` only — the `?TODO` near the
  `_interpreters.exec()` call points at anyio's
  `to_interpreter._interp_call` `(retval, is_exception)` pattern as
  the next step. Coordinates with the `trio.Cancelled` paths around
  `subint_exited.wait()`.

- [ ] Migrate SpawnSpec arg-passing to
  `_interpreters.set___main___attrs()`. `_subint.subint_proc()`'s
  `?TODO` at the `bootstrap` literal: same API anyio uses in
  `to_interpreter._Worker.call()`; needed once
  non-`repr()`-roundtrippable values (`SpawnSpec` struct,
  callables) get passed through.

- [ ] Implement `child_sigint='trio'` mode (or remove the flag).
  Scaffolded in `_main_thread_forkserver` but currently a no-op
  pending the orphan-SIGINT root-cause fix tracked in
  `ai/conc-anal/subint_forkserver_orphan_sigint_hang_issue.md`.
  Once trio's `epoll_wait` wedge is fixed, the flag may end up a
  no-op / doc-only mode.

- [ ] Add cancellation / hard-kill stress coverage for the
  forkserver backend (counterpart to
  `tests/spawn/test_subint_cancellation.py` for the plain `subint`
  backend). Module docstring lists this under "Still-open work".

- [ ] Run the `?TODO` typo-check enhancement in
  `tractor._testing.pytest` — pipe `skipon_spawn_backend` args
  through the try-set-backend checker rather than just `assert in
  get_args(SpawnMethodKey)`.

- [ ] xplatform pass for `tractor._testing._reap` — process-reap
  path is currently Linux-only via
  `/proc/<pid>/{status,cwd,cmdline}`; the `--shm` phase is
  Linux/FreeBSD only (macOS POSIX shm has no fs-visible path;
  Windows is a different story). Module docstring notes a
  `psutil`-based rewrite is viable since the dep is already
  test-time.

- [ ] Root-cause the **Mode-A cancel-cascade hang** under heavy
  fork-spawn contention (#451). Reproduces ~17% of runs at 3
  parallel pytest streams × `cpu_count - 2` actors; ≈0% on an
  idle single-stream system. Parent-side dump shows trio's main
  thread parked in `trio._core._io_epoll.get_events()` line 245
  — cancel cascade has reached the I/O wait but `epoll.poll`
  never returns. Workaround in this PR: per-test
  `trio.fail_after(12)` cap + `reap_subactors_per_test` opt-in
  fixture confine the failure to the originating test (no
  cascade contamination of sibling tests), so the suite stays
  green. Real fix needs a contention-amplifier reproducer +
  `stackscope` task-tree dumps from parent + each subactor at
  the t≈9 mid-cascade mark.

---

(this pr content was generated in some part by [`claude-code`][claude-code-gh])

[_main_thread_forkserver-py]: https://github.com/goodboy/tractor/blob/subint_forkserver_backend/tractor/spawn/_main_thread_forkserver.py
[_subint_forkserver-py]: https://github.com/goodboy/tractor/blob/subint_forkserver_backend/tractor/spawn/_subint_forkserver.py
[_subint-py]: https://github.com/goodboy/tractor/blob/subint_forkserver_backend/tractor/spawn/_subint.py
[_subint_fork-py]: https://github.com/goodboy/tractor/blob/subint_forkserver_backend/tractor/spawn/_subint_fork.py

[82332fbc]: https://github.com/goodboy/tractor/commit/82332fbc
[25e400d5]: https://github.com/goodboy/tractor/commit/25e400d5
[cf2e71d8]: https://github.com/goodboy/tractor/commit/cf2e71d8
[26914fde]: https://github.com/goodboy/tractor/commit/26914fde
[63ab7c98]: https://github.com/goodboy/tractor/commit/63ab7c98
[7804a9fe]: https://github.com/goodboy/tractor/commit/7804a9fe
[76605d56]: https://github.com/goodboy/tractor/commit/76605d56
[dcd5c1ff]: https://github.com/goodboy/tractor/commit/dcd5c1ff
[a72deef7]: https://github.com/goodboy/tractor/commit/a72deef7
[f5f37b69]: https://github.com/goodboy/tractor/commit/f5f37b69
[5e85f184]: https://github.com/goodboy/tractor/commit/5e85f184
[8bcbe730]: https://github.com/goodboy/tractor/commit/8bcbe730
[e31eb8d7]: https://github.com/goodboy/tractor/commit/e31eb8d7
[1e357dcf]: https://github.com/goodboy/tractor/commit/1e357dcf
[d093c319]: https://github.com/goodboy/tractor/commit/d093c319
[70d58c4b]: https://github.com/goodboy/tractor/commit/70d58c4b
[e3f4f5a3]: https://github.com/goodboy/tractor/commit/e3f4f5a3
[1af21210]: https://github.com/goodboy/tractor/commit/1af21210
[35da8089]: https://github.com/goodboy/tractor/commit/35da8089
[9993db01]: https://github.com/goodboy/tractor/commit/9993db01
[c20b05e1]: https://github.com/goodboy/tractor/commit/c20b05e1
[8ac3dfeb]: https://github.com/goodboy/tractor/commit/8ac3dfeb
[506617c6]: https://github.com/goodboy/tractor/commit/506617c6
[ab86f761]: https://github.com/goodboy/tractor/commit/ab86f761
[458a35cf]: https://github.com/goodboy/tractor/commit/458a35cf
[7cd47ef7]: https://github.com/goodboy/tractor/commit/7cd47ef7
[76d12060]: https://github.com/goodboy/tractor/commit/76d12060
[4106ba73]: https://github.com/goodboy/tractor/commit/4106ba73
[eceed29d]: https://github.com/goodboy/tractor/commit/eceed29d
[4c133ab5]: https://github.com/goodboy/tractor/commit/4c133ab5
[e312a68d]: https://github.com/goodboy/tractor/commit/e312a68d
[4d055543]: https://github.com/goodboy/tractor/commit/4d055543
[d6e70e9d]: https://github.com/goodboy/tractor/commit/d6e70e9d
[b350aa09]: https://github.com/goodboy/tractor/commit/b350aa09
[2ca0f41e]: https://github.com/goodboy/tractor/commit/2ca0f41e
[44bdb169]: https://github.com/goodboy/tractor/commit/44bdb169
[eae478f3]: https://github.com/goodboy/tractor/commit/eae478f3
[6d76b604]: https://github.com/goodboy/tractor/commit/6d76b604
[c99d475d]: https://github.com/goodboy/tractor/commit/c99d475d
[aa3e2309]: https://github.com/goodboy/tractor/commit/aa3e2309
[4f12d69b]: https://github.com/goodboy/tractor/commit/4f12d69b
[65fcfbf2]: https://github.com/goodboy/tractor/commit/65fcfbf2
[9b05f659]: https://github.com/goodboy/tractor/commit/9b05f659
[66f1941f]: https://github.com/goodboy/tractor/commit/66f1941f
[54561959]: https://github.com/goodboy/tractor/commit/54561959
[3ab99d55]: https://github.com/goodboy/tractor/commit/3ab99d55
[4b5176e2]: https://github.com/goodboy/tractor/commit/4b5176e2
[99dade0f]: https://github.com/goodboy/tractor/commit/99dade0f
[57dae0e4]: https://github.com/goodboy/tractor/commit/57dae0e4
[5e83881f]: https://github.com/goodboy/tractor/commit/5e83881f
[9f0709ee]: https://github.com/goodboy/tractor/commit/9f0709ee
[205382a3]: https://github.com/goodboy/tractor/commit/205382a3
[cbdf1eb6]: https://github.com/goodboy/tractor/commit/cbdf1eb6
[7c5dd4d0]: https://github.com/goodboy/tractor/commit/7c5dd4d0
[pr-446]: https://github.com/goodboy/tractor/pull/446


ai/conc-anal/subint_forkserver_thread_constraints_on_pep684_issue.md
```diff
@@ -0,0 +1,184 @@
+# Revisit `subint_forkserver` thread-cache constraints once msgspec PEP 684 support lands
+
+Follow-up tracker for cleanup work gated on the msgspec
+PEP 684 adoption upstream ([jcrist/msgspec#563](https://github.com/jcrist/msgspec/issues/563)).
```
goodboy, line 6: update this to the new subints compat issue we make @ `msgspec`.

ai/conc-anal/subint_forkserver_thread_constraints_on_pep684_issue.md
```diff
@@ -0,0 +1,184 @@
...
+
+Context — why this exists
```
goodboy, line 8: We can lose the `threading.Thread` once we can use *isolated (GIL) subints*..

ai/conc-anal/subint_forkserver_thread_constraints_on_pep684_issue.md
goodboy: https://github.com/jcrist/msgspec/issues/1026

tractor/spawn/_subint_forkserver.py
```diff
@@ -0,0 +1,907 @@
...
+in `tractor.spawn._spawn._methods` and selectable via
+`try_set_start_method('subint_forkserver')` / `--spawn-backend
+=subint_forkserver`. Parent-side spawn, child-side runtime
+bring-up and normal portal-RPC teardown are validated by the
+backend-tier test in
+`tests/spawn/test_subint_forkserver.py::test_subint_forkserver_spawn_basic`.
+
+Still-open work (tracked on tractor #379):
+
+- no cancellation / hard-kill stress coverage yet (counterpart
```
goodboy: Ah, we need to update all this content since the most recent version definitely covers almost all of these original outstanding issues.

tractor/spawn/_subint_forkserver.py
```diff
@@ -0,0 +1,907 @@
...
+  to `tests/test_subint_cancellation.py` for the plain
+  `subint` backend),
+- `child_sigint='trio'` mode (flag scaffolded below; default
```
goodboy: I think this should be a follow up tracking issue since generally speaking IPC-based graceful cancellation should be the desired default with a *`SIGINT`/`SIG<OTHER>`-to-each-subproc* as a timeout-driven "hard kill" backup method, but **only when enabled** by the backend's caller.

tractor/spawn/_subint_forkserver.py
```diff
@@ -0,0 +1,907 @@
...
+        _server,
+    )
+    from tractor.runtime._runtime import Actor
+    from tractor.runtime._supervise import ActorNursery
+
+
+log = get_logger('tractor')
+
+
+# Configurable child-side SIGINT handling for forkserver-spawned
```
goodboy: Still a todo and should have a follow up issue per comment above,

https://github.com/goodboy/tractor/pull/447/changes#r3150290471

tractor/spawn/_subint_forkserver.py
```diff
@@ -0,0 +1,1166 @@
...
+never entered a subint — works cleanly. Empirically validated
+across four scenarios by
+`ai/conc-anal/subint_fork_from_main_thread_smoketest.py` on
+py3.14.
+
+This submodule lifts the validated primitives out of the
+smoke-test and into tractor proper as the
+`subint_forkserver` spawn backend.
+
+Design rationale — why a forkserver, and why in-process
```
goodboy: Hopefully this very recently docs update will provide for most of the "explainer" for *why this design* 🙏🏼.

A further follow up question is whether we should break out the subint-free version of this current implementation (which is more accurately named something like a *fork-from-main-thread-not-running-`trio.run()`, where it's implied `trio.run()` is in a bg thread in each sub-actor.*)  into a diff module and leave this one to be filled out with the subint support once jcrist/msgspec#1026 unblocks supporting it?

tractor/_testing/pytest.py
```diff
@@ -24,16 +24,114 @@
...
+    from argparse import Namespace
+
+_cap_sys_passed_as_flag: bool = False
+_cap_fd_set: bool = False
+
+# XXX REQUIRED in order to enforce `--capture=` flag
+# pre test session.
+# https://docs.pytest.org/en/stable/reference/reference.html#bootstrapping-hooks
+@pytest.hookimpl(tryfirst=True)
+def pytest_load_initial_conftests(
```
goodboy, line 88: Why can't i force `--capture=sys` 😹

tractor/_testing/pytest.py
```diff
@@ -420,3 +600,85 @@ def pytest_generate_tests(
...
+) -> str:
+    if _is_forking_spawner(start_method):
+        request.getfixturevalue('capsys')
+        return 'sys'
+
+    return request.config.option.capture
+
+
+@pytest.fixture
+def set_fork_aware_capture(
```
goodboy, line 906: This dynamic override to `capsys` doesn't seem to actually work in the sense that the overall test run hangs if i don't **actually pass `--capture=sys` in the invocation cmd..

goodboy:
Superceded by #462 (surrounding tooling from the initial-and-WIP dev of), the newly factored MTF-backend-only #463.

PR #453, opened 2026-04-30
Auditing a new `main_thread_forkserver` spawning backend

goodboy:
Re-worked/squashed history from #447 but focused on promoting and mainlining the (nearly CI audited as working)  `tractor.spawn._main_thread_forkserver` backend tinkered from the histories (commit-chronologically) of,

-  #446
- #448
-  #447

Surprisingly (at least to me) this patch set was mostly vibe coded with **much** human intuition/guidance to get us to what seems to be a much superior approach to the stdlib's `multiprocessing` "forkserver" spawn method,

- https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
- https://docs.python.org/3/library/multiprocessing.html#the-spawn-and-forkserver-start-methods

which has been very problematic for us within rigorous "SC safe" constraints and more generally for nested subactor trees per,

- https://github.com/goodboy/tractor/issues/20
- https://github.com/goodboy/tractor/issues/84
- https://github.com/goodboy/tractor/issues/89
- https://github.com/goodboy/tractor/issues/134

---

#### Differences from #447 in this patch

despite the diff being identical this instead,
- focuses/emphasizes the 3.13+ and-already-working `.spawn._main_thread_forkerser` backend despite bringing in the code first prototyped in the `spawn._subint`, `spawn._subint_fork` and `spawn._subint_forkserver` backends each first presented in the draft PRs above.
- minimized set of larger commits reworked by `claude` to make the history by commit more immediately obvious, high-level-grokable for human onlookers.
- will ensure we get a clean CI run for all of,
  - [ ] `--spawn-backend=main_thread_forkserver` + `--tpt-proto=[tcp|uds]`
  - [ ] all prior cross-platform runs
  
- detail out various new `pytest` harness features developed while prototyping this backend and debugging various complex testing edge cases including some very difficult to decipher test-suite hangs on the new backend.

---

#### Still TODO before this lands,

- [ ] add new spawner-backend job set over TCP and UDS.
- [ ] bring in new test harness feats either here or in a pre-patch-PR.
- [ ] update the docs (at least the readme) to reflect the available backends more obviously; POSIX platforms are the only ones gaining from this spawner etc.


goodboy:
Heh, looking through these commits, i actually think they're a bit **too chunky** 😂 🫣 ..

so i might replace this with something else FYI. either way gonna use this one to get CI green..

goodboy:
Yeah, so much work since thinking this was near stable, and in hindsight i don't like losing the granular commit history from the very iterative work getting our test suite stable against this spawner.

Closing.

PR #479, opened 2026-07-02
Add `.trionics.start_or_cancel()` test suite

### LINKED ISSUE #474: We need test suite for `.trionics.start_or_cancel()`
Not sure how that got missed in #464 but definitely worth it to dig out the use case from `modden` as a unit test to include here (and provide motivation for upstream `trio` to possibly provide better porcelain around this case.

goodboy:
Resolve #474 with a new `tests/trionics/test_taskc.py` (9 tests) covering the `trio.Nursery.start()` wrapper landed in PR #464, incl. the `modden.runtime.progman.open_wks()` use case dug out as a minimal repro.

Deats,
- the lossy `RuntimeError('child exited without calling task_status.started()')` only fires when the child absorbs its ambient cancel pre-`.started()` (graceful-teardown pattern); a well-behaved child surfaces `Cancelled` direct from `.start()` on `trio` 0.29 - verified empirically 1st.
- `test_sibling_err_not_masked_by_startup_rte`: the `modden` case; ONLY the root-cause sibling `ValueError` escapes the nursery with the wrapper vs. bare-`.start()`'s lossy riding-along startup-RTE noise.
- `test_pure_oob_cancel_not_morphed_to_rte`: plain ancestor `cs.cancel()` exits clean vs. bare's eg-wrapped RTE.
- `test_genuine_startup_rte_still_raised`: sans cancellation the protocol-bug RTE re-raises same as bare.
- `test_childs_own_rte_never_demoted_to_cancel`: exact-msg + `isinstance`-guard regression cover; a child's own `RuntimeError('never got started!')`/`RuntimeError(1234)` never demotes to a `Cancelled`.
- `test_started_value_and_args_passthru`: positional args, `name=` and `.started()`-value forwarding.

Also,
- `use_start_or_cancel=False` params pin upstream `trio`'s current lossy behaviour as wart-documentation: a break on a `trio` upgrade likely means new upstream porcelain and the wrapper deserves a re-audit.
- verified 0 flakes over 50 hammer runs + 2 impl mutations each caught by exactly the targeted tests.

Prompt-IO: ai/prompt-io/claude/20260702T161624Z_65bf9df5_prompt_io.md (this patch was generated in some part by [`claude-code`][claude-code-gh]) [claude-code-gh]: https://github.com/anthropics/claude-code

### DIFF
--- a/ai/prompt-io/claude/20260702T161624Z_65bf9df5_prompt_io.md
+++ b/ai/prompt-io/claude/20260702T161624Z_65bf9df5_prompt_io.md
@@ -1,0 +1,80 @@
+---
+model: claude-fable-5
+service: claude
+session: f4da6654-63f9-4641-b567-f200da46d697
+timestamp: 2026-07-02T16:16:24Z
+git_ref: 65bf9df5
+scope: tests
+substantive: true
+raw_file: 20260702T161624Z_65bf9df5_prompt_io.raw.md
+---
+
+## Prompt
+
+> attempt to resolve
+> https://github.com/goodboy/tractor/issues/474
+> do it with /open-wkt.
+
+Issue #474: "We need test suite for
+`.trionics.start_or_cancel()`" — dig out the use case
+from `modden` as a unit test (and provide motivation
+for upstream `trio` to possibly provide better
+porcelain around this case). The wrapper landed via
+PR #464 with no tests.
+
+## Response summary
+
+New `tests/trionics/test_taskc.py` (9 tests) covering
+`start_or_cancel()`, authored in wkt
+`start_or_cancel_tests_474`.
+
+Behaviour was first probed empirically (trio 0.29):
+the lossy `RuntimeError('child exited without calling
+task_status.started()')` only fires when the child
+exits pre-`.started()` WITHOUT propagating the ambient
+`Cancelled` — i.e. when the child (or lib code it
+calls) absorbs the cancel in a graceful-teardown
+pattern; a well-behaved child surfaces `Cancelled`
+straight out of `.start()`. The `modden`
+`progman.open_wks()` use case was reconstructed from
+`modden/runtime/progman.py` accordingly.
+
+Tests (each `use_start_or_cancel` parametrization also
+pins upstream trio's current lossy behaviour as
+wart-documentation):
+
+- `test_sibling_err_not_masked_by_startup_rte` — the
+  `modden` case: sibling error OOB-cancels the shared
+  nursery scope; with the wrapper ONLY the root-cause
+  `ValueError` escapes; bare `.start()` adds the lossy
+  RTE alongside.
+- `test_pure_oob_cancel_not_morphed_to_rte` — plain
+  ancestor `cs.cancel()`: wrapper → clean exit; bare
+  → eg-wrapped RTE.
+- `test_genuine_startup_rte_still_raised` — no
+  cancellation → protocol-bug RTE re-raised same as
+  bare.
+- `test_childs_own_rte_never_demoted_to_cancel` — a
+  child's own `RuntimeError('never got started!')` /
+  `RuntimeError(1234)` under ambient cancel is never
+  demoted to `Cancelled` (exact-msg-match + str-guard
+  regression cover).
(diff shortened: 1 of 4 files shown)


goodboy (commented):
### [P2] Assert cancellation escapes the wrapper call itself

`tests/trionics/test_taskc.py:93-115,146-168` | confidence: high | category: correctness

The two core cancellation tests inspect only the eventual nursery/scope outcome. They would still pass if `start_or_cancel()` caught Trio’s startup `RuntimeError` and simply returned `None` instead of re-raising the ambient cancellation:

- The sibling-error case would still expose only the sibling `ValueError`.
- The pure-cancel case could observe cancellation later during nursery exit.

Neither proves that `checkpoint_if_cancelled()` interrupts the exact `await start_or_cancel(...)` boundary.

Recommendation: catch `trio.Cancelled` immediately around that await, record that the wrapper raised it, then re-raise and assert the record after leaving the cancelled scope.

### [P2] Exercise the actual over-broad startup-error predicate

`tests/trionics/test_taskc.py:224-275` | confidence: high | category: correctness

The test claims to protect an “exact-message guard,” but its string case is `"never got started!"`. The implementation actually uses:

```python
'child exited without calling' in rte.args[0]
```

A child-owned error such as `RuntimeError("child exited without calling user hook")` would therefore be mistaken for Trio’s startup error and demoted to `Cancelled` under ambient cancellation. The new suite would still pass.

Recommendation: add a child-owned message containing that exact prefix but differing from Trio’s canonical full message, then tighten the implementation to compare against the complete canonical wording.

### [P3] Replace the two-second ordering timer

`tests/trionics/test_taskc.py:37-52` | confidence: medium | category: reliability

`absorbs_cancel_pre_started()` uses `trio.sleep(2)` to keep the child before `.started()`. Under a sufficiently delayed or instrumented CI process, that deadline can become runnable before the canceller task executes, allowing the startup path to win and making the cancellation tests flaky.

Recommendation: use `trio.sleep_forever()` for the cancellation-absorbing child. If readiness ordering matters, coordinate it with `trio.Event` and apply a timeout around the complete scenario rather than inside the child.

### [P3] Verify the child error propagates unchanged

`tests/trionics/test_taskc.py:248-275` | confidence: high | category: correctness

The test says the child’s `RuntimeError` propagates unchanged, but checks only the first grouped exception’s type and argument. Replacing it with a newly constructed error, or including extra leaves, would pass.

Recommendation: pre-create the exception, raise that object, assert identity, and assert the group contains exactly that leaf.

**Checks skipped**
- Tests were not run because the review request did not authorize execution.
- Static analyzers were not run.
- Prompt-IO files were treated as generated provenance and excluded from semantic review.

**Scope**
- Local/prospective review for PR `#479`
- Base tip and merge base: `92c737ad`
- Head: `203a0f7e`
- Range: `main...HEAD`
- Semantic path: `tests/trionics/test_taskc.py`
- Generated exclusions: two `ai/prompt-io/claude/` artifacts

goodboy (commented):
### Review findings resolved

All four findings from review
[`4920098929`](https://github.com/goodboy/tractor/pull/479#pullrequestreview-4920098929)
are addressed:

- [`1c7d0c7f`](https://github.com/goodboy/tractor/commit/1c7d0c7f3ea3aaa5bdd0262e4553f80464ef23bb)
  matches Trio's complete canonical startup error and verifies that
  colliding child-owned `RuntimeError` objects remain the sole
  exception-group leaf with identity preserved.

- [`ae67e2f4`](https://github.com/goodboy/tractor/commit/ae67e2f4296801309079dda728fa73ce25d3cfb0)
  replaces the wall-clock ordering timer with `sleep_forever()` and
  asserts that `Cancelled` escapes each exact `start_or_cancel()`
  await before nursery or cancel-scope handling.

Verification at `ae67e2f4`:

- `ruff check` on `_taskc.py` and `test_taskc.py`
- `10 passed` in `tests/trionics/test_taskc.py`
- all GitHub CI jobs pass
- PR is `MERGEABLE` and `CLEAN`

(this review was generated in some part by `opencode` using
`gpt-5.6-sol` (`openai`))

Issue #473, opened 2026-06-29
Root-cause UDS-on-macOS — `uds_transport_actor_tree` example fails on CI


goodboy:
## Summary

The new `examples/uds_transport_actor_tree.py` (added in #460,
`759d46dd`) **reliably fails on macOS CI** (`returncode 1`), while
passing cleanly on Linux. It's skipped on macOS+CI in #460 (so the
docs PR lands green) — this issue tracks the real root-cause.

## What we know

- The example forces `enable_transports=['uds']`.
- UDS-on-macOS is **otherwise un-exercised by the CI matrix** —
  there's only one macOS job (`tpt_proto=tcp`), no `tpt_proto=uds`
  macOS job. So this example is the first thing to surface a macOS
  UDS-path issue.
- The peer-cred paths are already platform-guarded + graceful
  (`get_peer_info`/`SO_PEERCRED` on Linux vs
  `get_peer_pid`/`LOCAL_PEERPID` elsewhere, the latter returns `None`
  on failure), so the crash is **deeper in the macOS UDS stack**.
- `tests/test_docs_examples.py` swallows the subproc stderr (only
  re-raises when the last line contains `"Error"`), so the exact
  macOS failure can't be read from CI logs as-is.

## Next steps

- [x] Surface the subproc stderr on any non-zero exit in
  `test_docs_examples.py` (diagnostic patch on a branch) so a single
  macOS CI run reveals the real traceback.
- [x] Root-cause the macOS UDS failure from that traceback.
- [x] Add a `tpt_proto=uds` macOS job to the matrix so UDS-on-macOS
  is actually covered going forward.
- [x] Un-skip the example on macOS once fixed.

(this issue was generated in some part by [`claude-code`][cc])

[cc]: https://github.com/anthropics/claude-code

goodboy:
Resolved by #480 💥

PR #460, opened 2026-06-11
Docs (vibed) for serious: real documentation with a modern theme 

goodboy:
### Motivation

Our docs were years out of date — teaching a long-dead API
(`tractor.run()`, `@stream`, "arbiters") on a barebones theme — and
#157 had been asking for a proper modern docs site for ages, with
#175/#126/#123 piling on tutorial/testing/deploy asks.

This rebuilds the docs from the ground up: a 3-column
`pydata-sphinx-theme` site skinned black-and-white, `d2`-rendered
diagrams, and prose that teaches ONLY the current API — with every
code block `literalinclude`-d from `examples/` so the docs can't rot
(what you read is what CI runs).

### Src of research

Theme + tooling choices were research-backed (the #157 thread plus a
live survey of peer projects),

- `pydata-sphinx-theme` — numpy / ray / polars all use it today (ray
  migrated off `sphinx-book-theme`).
- `d2` (terrastruct) for diagrams-as-code.
  * https://d2lang.com

### Summary of changes

- Stand up the sphinx project on `pydata-sphinx-theme` 0.18 + sphinx
  9: new `docs/conf.py`, a `docs` dependency-group, a minimal b&w
  `custom.css`, and an opt-in `nix develop .#docs` shell carrying the
  `d2` renderer.
- Add a local `.. d2::` directive (`docs/_ext/d2diagrams.py`) that
  renders `d2` sources at build time with a committed-SVG fallback
  (CI needs no `d2` bin) + a vendored `.. margin::`, plus 7
  hand-authored diagrams.
- Write the content tree — landing, an "SC across processes"
  explainer + runtime architecture, 12 task guides (the flagship
  multi-process debugger, `Context`/streaming, cancellation,
  discovery, infected `asyncio`, typed msging, testing) and 10
  curated autodoc API pages — and fix the lib docstrings'
  informal-RST so the build is warning-free.
- Add/modernize the examples backing the guides: 5 new gap-filling
  scripts, the dated set moved to `.wait_for_result()`, and the
  stdlib primes baseline un-hidden — all CI-run.
- Redesign the landing hero as an inline-svg wireframe that flips
  with the light/dark toggle (`fill: currentColor`), carry it to the
  navbar + README, and brand the navbar polars-style.
- Wire up GitHub Pages: a build-on-PR / deploy-on-`main` workflow,
  `sphinx.ext.githubpages` (`.nojekyll`) + `html_baseurl`.
[...]

PR #480, opened 2026-07-02
Fix `uds` addr corruption + exercise macOS CI

goodboy:
### Motivation

The new `examples/uds_transport_actor_tree.py` (PR #460) reliably
fails on macOS CI but the docs-example harness swallows the
traceback: it only re-raises captured stderr when the LAST line
contains 'Error', and a `tractor` root-actor crash always ends with
the strict-EG collapse note `( ^^^ this exc was collapsed from a
group ^^^ )` — so every possible crash reduces to a bare `assert 1 ==
0` in the CI logs. Worse, UDS-on-macOS is otherwise entirely
un-exercised: the test matrix explicitly excludes the `macos-latest`
+ `uds` combo, so the (skipped) example was the only thing that ever
touched that path.

This branch lands the diagnosis plan from GH #473: surface the real
stderr, fix the one macOS UDS bug provable from linux, and turn on
the missing CI coverage so the next macOS run either goes green or
pins any remaining crasher layer with a full traceback.

---

### Summary of changes

- `tests.test_docs_examples`: always raise with the FULL subproc
  stderr (+stdout) whenever an example exits non-zero, regardless of
  stderr shape; un-skip the `uds` example on macOS+CI.
- `tractor.ipc._uds`: fix accept-side addr corruption on no-autobind
  platforms (macOS!) — the `(str, str)` match arm in
  `MsgpackUDSStream.get_stream_addrs()` took `getpeername()`
  unconditionally, which is `''` on the accept side without linux's
  `SO_PASSCRED`-triggered autobind, corrupting every accepted conn's
  laddr/raddr to `Path('')`. Proven + verified via a linux sim
  (nulling `SO_PASSCRED`).
  * also harden the bindspace `mkdir()` w/ `parents=True,
    exist_ok=True` (nested custom `filedir`s + racing actors).
- `.github/workflows/ci.yml`: drop the matrix `exclude` so the full
  suite runs with `--tpt-proto=uds` on `macos-latest`.
- `ai/prompt-io/`: session provenance log per the NLNet generative-AI
  policy.

---

### TODOs before landing

- [ ] Confirm the new `macos-latest` + `uds` CI job and the
  un-skipped example: the addr fix is non-fatal on the linux sim so
  it may not be the (only) macOS crasher — if the leg is still red,
  the now-surfaced traceback pins the remaining layer (which is
  itself the GH #473 deliverable).

---

### Future follow up

Tracked in [GH #473][473] — its remaining checkboxes get checked off
with commit ref-links as part of landing this.

(this pr content was generated in some part by [`claude-code`][claude-code-gh])

[473]: https://github.com/goodboy/tractor/issues/473

### DIFF
--- a/tests/discovery/test_daemon_fixture.py
+++ b/tests/discovery/test_daemon_fixture.py
@@ -1,6 +1,10 @@
 '''
 Discovery daemon fixture regressions.
 
+This module imports private helpers from the sibling
+`tests.discovery.conftest` plugin to exercise that fixture machinery
+directly, rather than testing a production `tractor` API.
+
 '''
 from unittest.mock import (
     call,
--- a/tests/discovery/test_daemon_fixture.py
+++ b/tests/discovery/test_daemon_fixture.py
@@ -15,7 +15,7 @@ def test_daemon_ready_check_does_not_connect(
     tmp_path,
 ):
     '''
-    Detect a listening UDS daemon without creating a raw connection.
+    Observe completed daemon startup without a raw connection.
 
     The old UDS readiness helper connected and immediately closed. That
     entered Tractor's actor-handshake handler with no `Aid` payload and
--- a/tests/discovery/test_daemon_fixture.py
+++ b/tests/discovery/test_daemon_fixture.py
@@ -1,0 +1,72 @@
+'''
+Discovery daemon fixture regressions.
+
+'''
+from unittest.mock import (
+    call,
+    Mock,
+)
+
+from .conftest import _wait_for_daemon_ready
+
+
+def test_daemon_ready_check_does_not_connect(
+    monkeypatch,
+    tmp_path,
+):
+    '''
+    Detect a listening UDS daemon without creating a raw connection.
+
+    The old UDS readiness helper connected and immediately closed. That
+    entered Tractor's actor-handshake handler with no `Aid` payload and
+    destabilized the remote registrar on macOS before discovery tests
+    started. This test creates the child sentinel, forbids all socket
+    construction and connection helpers, then proves readiness returns
+    without touching the transport layer.
+
+    '''
+    ready_path = tmp_path / 'daemon-ready'
+    ready_path.touch()
+    socket_ctor = Mock(side_effect=AssertionError('socket opened'))
+    connect = Mock(side_effect=AssertionError('socket connected'))
+    monkeypatch.setattr('socket.socket', socket_ctor)
+    monkeypatch.setattr('socket.create_connection', connect)
+
+    _wait_for_daemon_ready(
+        ready_path=ready_path,
+        deadline=.1,
+        poll_interval=.01,
+    )
+
+    socket_ctor.assert_not_called()
+    connect.assert_not_called()
+
+
+def test_daemon_ready_check_backs_off(monkeypatch):
+    '''
+    Back off while waiting for the child startup sentinel.
+
+    The sentinel may appear after several parent polling intervals.
+    A deterministic false/false/true path sequence proves the helper
+    sleeps between unsuccessful observations instead of hot-spinning
+    and starving a booting daemon on constrained CI workers.
+
+    '''
+    ready_path = Mock()
+    ready_path.is_file.side_effect = [False, False, True]
+    sleep = Mock()
+    monotonic = Mock(side_effect=[0, 0, 0, 0])
+    monkeypatch.setattr('time.sleep', sleep)
+    monkeypatch.setattr('time.monotonic', monotonic)
+
+    _wait_for_daemon_ready(
+        ready_path=ready_path,
+        deadline=.2,
+        poll_interval=.01,
+    )
+
+    assert ready_path.is_file.call_count == 3
+    assert sleep.call_args_list == [
+        call(.01),
+        call(.01),
+    ]
--- a/tests/test_docs_examples.py
+++ b/tests/test_docs_examples.py
@@ -145,21 +145,6 @@ def test_example(
             'This test does run just fine "in person" however..'
         )
 
-    if (
-        'uds_transport_actor_tree' in ex_file
-        and
-        _friggin_macos
-        and
-        ci_env
-    ):
-        pytest.skip(
-            'UDS-transport example reliably fails on macOS CI.\n'
-            'UDS-on-macOS is otherwise un-exercised by the matrix\n'
-            '(no `tpt_proto=uds` macOS job), so this new example is\n'
-            'the first to surface it; the macOS UDS path needs\n'
-            'root-causing. Passes on Linux.'
-        )
-
     from .conftest import cpu_perf_headroom
 
     timeout: float = (
--- a/tests/test_docs_examples.py
+++ b/tests/test_docs_examples.py
@@ -178,10 +178,11 @@ def test_example(
         code = ex.read()
 
         with run_example_in_subproc(code) as proc:
+            out = None
             err = None
             try:
                 if not proc.poll():
-                    _, err = proc.communicate(timeout=timeout)
+                    out, err = proc.communicate(timeout=timeout)
 
             except subprocess.TimeoutExpired as e:
                 test_log.exception(
@@ -190,9 +191,34 @@ def test_example(
                 proc.kill()
                 err = e.stderr
 
+            errmsg: str = err.decode() if err else ''
+
+            # XXX, ALWAYS surface the subproc's full stderr
+            # whenever it exits non-zero!
+            #
+            # The prior impl only raised when the LAST stderr
+            # line contained 'Error', swallowing any crash whose
+            # traceback ends in a non-`XxxError:` line; in
+            # particular EVERY `tractor` root-actor crash ends
+            # with the strict-EG collapse note,
+            # '( ^^^ this exc was collapsed from a group ^^^ )',
+            # so ALL such failures were reduced to a bare
+            # `assert 1 == 0` in CI logs.. see GH #473.
+            rc: int|None = proc.returncode
+            if rc:
+                outmsg: str = out.decode() if out else ''
+                raise Exception(
+                    f'Example script exited with rc={rc} !?\n'
+                    f'\n'
+                    f'stdout:\n'
+                    f'{outmsg}\n'
+                    f'\n'
+                    f'stderr:\n'
+                    f'{errmsg}\n'
+                )
+
             # if we get some gnarly output let's aggregate and raise
-            if err:
-                errmsg = err.decode()
+            if errmsg:
                 errlines = errmsg.splitlines()
                 last_error = errlines[-1]
                 if (
--- a/tests/test_docs_examples.py
+++ b/tests/test_docs_examples.py
@@ -48,6 +48,8 @@ def _reap_killed_proc(
         return proc.communicate()
 
     proc.wait(timeout=5)
+    if proc.stdin:
+        proc.stdin.close()
     if proc.stdout:
         proc.stdout.close()
     if proc.stderr:
@@ -257,7 +259,15 @@ def run_example_in_subproc(
         assert not proc.returncode
         try:
             yield proc
-        finally:
+        except BaseException:
+            if proc.poll() is None:
+                try:
+                    _kill_proc_tree(proc)
+                    _reap_killed_proc(proc)
+                except Exception:
+                    pass
+            raise
+        else:
             if proc.poll() is None:
                 _kill_proc_tree(proc)
                 _reap_killed_proc(proc)
--- a/tests/test_docs_examples.py
+++ b/tests/test_docs_examples.py
@@ -5,11 +5,13 @@ Let's make sure them docs work yah?
 from contextlib import contextmanager
 import itertools
 import os
+import signal
 import sys
 import subprocess
 import platform
 import shutil
 from typing import Callable
+from unittest.mock import Mock
 
 import pytest
 import tractor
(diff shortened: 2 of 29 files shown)


goodboy (commented):
### [P2] Shorten the runtime bindspace before enabling macOS UDS

`.github/workflows/ci.yml:117-120` | confidence: high | category: reliability

The patch enables the macOS UDS matrix and removes the example skip, but macOS limits Unix-domain socket paths more strictly than Linux. The default pytest runtime path exceeds that limit:

```text
/private/var/folders/.../pytest-of-runner/pytest-0/test_example10/.../tractor/registry@1616.sock
OSError: AF_UNIX path too long
```

This already caused `test_example[uds_transport_actor_tree.py]` to fail in the prior macOS TCP job. `start_listener()` still binds the complete path without checking or shortening it, so enabling the new required leg does not make macOS UDS viable.

Recommendation: allocate the macOS runtime bindspace beneath a predictably short directory, such as a compact `/tmp` path, or fail early with a clear path-length error. Add a regression test exercising the maximum supported path.

### [P2] Capture output from processes that have already failed

`tests/test_docs_examples.py:165-170` | confidence: high | category: diagnostics

`proc.poll()` returns the non-zero exit status when a process fails quickly. Because the code calls `communicate()` only under `if not proc.poll()`, those failures skip output collection entirely. The new exception then reports empty stdout and stderr, defeating the diagnostic improvement this patch introduces.

Recommendation: call `proc.communicate(timeout=timeout)` unconditionally. It safely collects output whether the process is running or has already exited.

### [P2] Reap the subprocess and its actor descendants after timeout

`tests/test_docs_examples.py:172-177` | confidence: high | category: reliability

On timeout, the code kills only the immediate example process and does not call `communicate()` or `wait()` afterward. The later assertion can escape through the context manager’s `yield`, bypassing its trailing `proc.wait()`. This leaves the process unreaped, while actor descendants may continue holding inherited stdout/stderr pipes.

That failure mode is consistent with the historical macOS UDS job reaching 89% and then remaining active until the 16-minute workflow timeout.

Recommendation: launch POSIX examples in a separate process group, terminate the group on timeout, and always drain/reap the immediate process in a `finally` block.

**Open Questions**

None affecting these findings.

**Checks Run**

- Inspected the complete `main...HEAD` diff and surrounding call paths.
- Ran `git diff --check`; no whitespace errors.
- Inspected historical macOS TCP and UDS logs.
- Queried fresh PR checks at head `451e0acf8a754f0323a1182f2894433316325802`; macOS and Ubuntu matrix jobs remain in progress.

**Checks Not Run**

- No local tests or analyzers were run during this read-only review.
- Fresh CI had not completed when this report was finalized.

**Scope**

Local/prospective review of PR `#480`.

- Base tip and merge base: `7554f90e5937dc9f7334a7c2179a8f991780fe82`
- Head: `451e0acf8a754f0323a1182f2894433316325802`
- Range: `main...HEAD`
- Reviewed paths: `.github/workflows/ci.yml`, `examples/uds_transport_actor_tree.py`, `tests/test_docs_examples.py`, `tractor/ipc/_uds.py`
- Prompt provenance files were excluded as non-runtime artifacts.

(this review was generated in some part by `OpenCode` using `gpt-5.6-sol` (`OpenAI`))

tests/test_docs_examples.py
```diff
@@ -178,10 +163,11 @@ def test_example(
         code = ex.read()
 
         with run_example_in_subproc(code) as proc:
+            out = None
             err = None
             try:
                 if not proc.poll():
-                    _, err = proc.communicate(timeout=timeout)
+                    out, err = proc.communicate(timeout=timeout)
```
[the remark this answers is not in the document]
goodboy: > response authored by `opencode`

The harness now calls `communicate()` unconditionally, including for fast non-zero exits, so both stdout and stderr are retained.

> 📎 fixed in [`0a3b0efc`](https://github.com/goodboy/tractor/commit/0a3b0efcc6af2e9b107d106ce4c78260bf620024)

tests/test_docs_examples.py
```diff
@@ -190,9 +176,34 @@ def test_example(
                 proc.kill()
```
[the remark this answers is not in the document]
goodboy: > response authored by `opencode`

POSIX examples now run in dedicated sessions. Timeout cleanup terminates the process group, then drains and reaps the leader before raising diagnostics.

> 📎 fixed in [`0a3b0efc`](https://github.com/goodboy/tractor/commit/0a3b0efcc6af2e9b107d106ce4c78260bf620024)

tests/test_docs_examples.py
```diff
@@ -190,9 +176,34 @@ def test_example(
...
                 err = e.stderr
 
+            errmsg: str = err.decode() if err else ''
```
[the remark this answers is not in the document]
goodboy: > response authored by `opencode`

Captured stderr now uses `errors='replace'`, preserving the original subprocess failure when output contains invalid bytes.

> 📎 fixed in [`0a3b0efc`](https://github.com/goodboy/tractor/commit/0a3b0efcc6af2e9b107d106ce4c78260bf620024)

tests/test_docs_examples.py
```diff
@@ -190,9 +176,34 @@ def test_example(
...
+            # line contained 'Error', swallowing any crash whose
+            # traceback ends in a non-`XxxError:` line; in
+            # particular EVERY `tractor` root-actor crash ends
+            # with the strict-EG collapse note,
+            # '( ^^^ this exc was collapsed from a group ^^^ )',
+            # so ALL such failures were reduced to a bare
+            # `assert 1 == 0` in CI logs.. see GH #473.
+            rc: int|None = proc.returncode
+            if rc:
+                outmsg: str = out.decode() if out else ''
```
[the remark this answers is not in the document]
goodboy: > response authored by `opencode`

Captured stdout now also uses `errors='replace'`, so malformed output cannot mask the process failure.

> 📎 fixed in [`0a3b0efc`](https://github.com/goodboy/tractor/commit/0a3b0efcc6af2e9b107d106ce4c78260bf620024)

tests/test_docs_examples.py
```diff
@@ -190,9 +176,34 @@ def test_example(
...
+                raise Exception(
+                    f'Example script exited with rc={rc} !?\n'
+                    f'\n'
+                    f'stdout:\n'
+                    f'{outmsg}\n'
+                    f'\n'
+                    f'stderr:\n'
+                    f'{errmsg}\n'
+                )
```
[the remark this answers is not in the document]
goodboy: > response authored by `opencode`

Fixture cleanup now suppresses teardown failures only while preserving an exception already raised by the test body. It no longer asserts over detailed stdout/stderr diagnostics.

> 📎 fixed in [`bd38204f`](https://github.com/goodboy/tractor/commit/bd38204fde928bd26307d29c81a5cf7e2bd9d0d1)

goodboy (commented):
> response authored by `opencode`

All three findings are addressed:

- Darwin uses a short, owner-only runtime bindspace, with regressions covering the 104-byte socket-path budget and generated-name compaction.
- Fast non-zero examples unconditionally drain stdout and stderr with non-masking decoding.
- Timed-out POSIX examples run in dedicated sessions; cleanup kills the process group, drains the pipes, and reaps the leader without masking a body failure.

> 📎 fixed in [`340d5069`](https://github.com/goodboy/tractor/commit/340d5069403404f665764a2551d56b97423cada2) [`0a3b0efc`](https://github.com/goodboy/tractor/commit/0a3b0efcc6af2e9b107d106ce4c78260bf620024) [`bd38204f`](https://github.com/goodboy/tractor/commit/bd38204fde928bd26307d29c81a5cf7e2bd9d0d1)

tests/discovery/test_daemon_fixture.py
```diff
@@ -0,0 +1,72 @@
+'''
+Discovery daemon fixture regressions.
+
```
goodboy, line 7: explain that this tests the `tests/discovery/conftest` fixture itself maybe?

tractor/runtime/_state.py
```diff
@@ -361,13 +401,46 @@ def get_rt_dir(
...
+    if rt_root is None:
+        if subdir_path is not None:
+            rt_dir = rt_dir / subdir_path
+        if not rt_dir.is_dir():
+            rt_dir.mkdir(
+                # Runtime dirs hold IPC sockets; owner-only access
+                # prevents other users from traversing the bindspace.
+                mode=0o700,
+                parents=True,
+                exist_ok=True,
```
goodboy, line 444: wait so if it already exists that's allowed, but then we're not enforcing owner-only-access right?

tractor/runtime/_state.py
```diff
@@ -361,13 +401,46 @@ def get_rt_dir(
...
+                or
+                dir_stat.st_uid != os.getuid()
+            ):
+                raise PermissionError(
+                    f'Unsafe Darwin runtime directory!\n'
+                    f'path: {rt_dir}\n'
+                    f'owner uid: {dir_stat.st_uid}\n'
+                    f'mode: {stat.filemode(dir_stat.st_mode)}\n'
+                )
+            if stat.S_IMODE(dir_stat.st_mode) != 0o700:
```
goodboy: also here, aren't we forcing the ownership bits even when the dir-path might be pre-existing?

tractor/_root.py
```diff
@@ -83,6 +86,68 @@
 )
 
 
+async def _probe_registry(
```
goodboy: maybe this should go under `.discovery.*` sugpkg?

tractor/_root.py
```diff
@@ -453,32 +518,28 @@ async def open_root_actor(
 
         # closed into below ping task-func
         ponged_addrs: list[Address] = []
+        occupied_addrs: list[Address] = []
 
         async def ping_tpt_socket(
```
goodboy: maybe this should be moved to `.discovery.*` as well?

tractor/ipc/_transport.py
```diff
@@ -61,6 +62,33 @@
 log = get_logger()
 
 
+def _peer_closed_errno(exc: BaseException) -> int|None:
+    '''
+    Find a peer-close errno in a transport exception chain.
+
+    '''
+    seen: set[int] = set()
+    while (
```
goodboy: just to clarify, this will iterate [`ExceptionGroup.exceptions`](https://docs.python.org/3.13/library/exceptions.html#BaseExceptionGroup.exceptions) implicitly then right?

goodboy (commented):
Actually **really happy** with how this turned out despite how tedious it was 🥳 

We now have an opsec follow up in #494 to get root-actor level (at least) perms guarantees around UDS bindspaces in the future 🏄🏼

tests/discovery/test_daemon_fixture.py
goodboy: > response authored by `opencode`

Yep. The module docstring now explains that this file imports private helpers from the sibling `tests.discovery.conftest` plugin and directly regression-tests that fixture machinery rather than a production API.

> 📎 fixed in [`75cda193`](https://github.com/goodboy/tractor/commit/75cda1933c98e6eefca550393c483bc5fcb9ec13)

tractor/runtime/_state.py
goodboy: > response authored by `opencode`

Darwin already rejected non-directories and foreign UIDs before normalizing mode to `0o700`, but good catch: Linux did not enforce that contract for an existing runtime directory. The shared POSIX guard now applies the same owner/type/mode checks on Linux. Process-level root-actor ownership is a separate boundary tracked in #494, including Linux mount-namespace research.

> 📎 fixed in [`f454cefe`](https://github.com/goodboy/tractor/commit/f454cefe56f96b502f8b0c754b6dfe977851b2c3)

tractor/runtime/_state.py
goodboy: > response authored by `opencode`

Yes, the mode is intentionally normalized for a pre-existing managed bindspace, but only after `lstat()` proves it is a directory owned by the current UID. Foreign-owned paths and symlinks are rejected before `chmod`, with Linux regressions covering both mode repair and rejection without mutation.

> 📎 fixed in [`f454cefe`](https://github.com/goodboy/tractor/commit/f454cefe56f96b502f8b0c754b6dfe977851b2c3)

tractor/_root.py
goodboy: > response authored by `opencode`

Agreed. The bounded single-address registrar handshake probe now lives in `tractor.discovery._api` as `_probe_registry()`; `_root` only consumes its discovery result.

> 📎 fixed in [`75cda193`](https://github.com/goodboy/tractor/commit/75cda1933c98e6eefca550393c483bc5fcb9ec13)

tractor/_root.py
goodboy: > response authored by `opencode`

Agreed. The concurrent multi-address wrapper moved with the single-address probe as `discovery._api._probe_registry_addrs()`, leaving `open_root_actor()` to consume the classified registrar and occupied-address lists.

> 📎 fixed in [`75cda193`](https://github.com/goodboy/tractor/commit/75cda1933c98e6eefca550393c483bc5fcb9ec13)

tractor/ipc/_transport.py
goodboy: > response authored by `opencode`

No, the previous cause/context loop did not traverse `BaseExceptionGroup.exceptions` implicitly. It now classifies grouped branches explicitly and normalizes only when every branch resolves to `EPIPE` or `ECONNRESET`; a mixed group preserves the original `trio.BrokenResourceError` instead of hiding an unrelated failure.

> 📎 fixed in [`f75c9cde`](https://github.com/goodboy/tractor/commit/f75c9cdeab681db5b576cb972bb14be9150dab5b)

PR #481, opened 2026-07-02
Add `tractor.to_actor` one-shot task API subpkg

goodboy:
### Motivation

Issue #477 asks us to decouple one-shot remote-task invocation from
`ActorNursery.run_in_actor()` and its secondary nursery/reaper
machinery. Following the `trio.to_thread` / `anyio.to_process`
parlance, this PR adds `tractor.to_actor.run()`: a blocking,
caller-scoped API built from daemon spawning, linked contexts and
deterministic process reaping.

Results and boxed remote errors now arrive in the calling task, so
ordinary local `trio` scopes own collection and cancellation instead
of deferring failures to actor-nursery teardown.

### Summary of changes

- add `tractor.to_actor.run(fn, *args, ...) -> RetT`. Target arguments
  follow Trio's positional convention; bind target keyword arguments
  with `functools.partial()`. Calls are normalized and signature-bound
  before runtime startup, including nested partials and Python 3.14
  `functools.Placeholder` semantics.
- run the target through a private `@context` trampoline and
  `Portal.open_context()`, preserving linked cancellation and the
  actor's RPC module allowlist. Reused actors must expose both the
  target module and `tractor.to_actor.MODULE`.
- support three actor-lifetime modes: no handle opens a private
  call-scoped actor nursery and implicit runtime; `an=` creates a
  fresh child in a caller-managed actor nursery; `portal=` reuses an
  existing actor and leaves that actor's lifetime with its owner.
  `runtime_kwargs` applies only to the private-runtime mode.
- always cancel, join and unregister call-owned children before
  returning. A missing cancel acknowledgement escalates to
  `proc.kill()`, while debugger-held trees retain their REPL
  protection.
- close actor-nursery registration/reaping races by publishing child
  and reap state atomically, making late registrations observe
  nursery cancellation, and retaining reap requests across concurrent
  startup and teardown.
- bound actor and context cancellation request publication and
  acknowledgement with one absolute deadline. IPC frame writes remain
  shielded until complete; an explicit deadline that catches a
  partial frame destroys the stream before releasing the send lock,
  while pending caller cancellation retains precedence over secondary
  transport-close errors.
- add dedicated API coverage for all lifetime modes, implicit runtime
  teardown, argument normalization and validation, module exposure,
  linked task cancellation, direct error relay, deterministic hard
  reaping and late-registration races. Extend context, transport,
  debugger, discovery and namespace-path regression coverage.
- document the final API throughout the API reference, quickstart and
[...]

PR #484, opened 2026-07-06
Drop `run_in_actor()` + the ria reap-cluster

### LINKED ISSUE #477: Factor `ActorNursery.run_in_actor()` (and possibly `Portal.run()`) into a new `tractor.to_actor` wrapper API
Originally from the discussion in #287,
- https://github.com/goodboy/tractor/issues/287#issuecomment-1005251485
- https://github.com/goodboy/tractor/issues/287#issuecomment-1005746676
- https://github.com/goodboy/tractor/issues/287#issuecomment-1005987203

> Yeah, after spelunking the original design a bit yesterday I *think* I'm of the opinion that we should *at the least* decouple `.run_in_actor()` from the nursery internals and *at most* possibly deprecate it from the nursery API.
> 
> 
> 
> Argument:
> 
> - `.run_in_actor()` calls can be implemented using `.start_actor()` + a `Portal.run()` call spawned from a local `trio` nursery task

further, a `Portal.run()` call can be implemented using `Portal.open_context()` with special handling on child side in the RPC `._invoke()` impl; it might actually already be done that way?
 
> -> we can offer a "wrapper" high level API that provides `.run_in_actor()` as part of a small worker pool API?

I like adopting the parlance from other projects such as `.to_thread`, `to_process`, and think we should instead offer a `tractor.to_actor`. subpkg of APIs which deliver these "higher level one shot" style apis for doing single remote task invocation.

> The benefits include:
> 
> - massively simplifying our actor nursery cancellation/spawn machinery
> 
> - moving error collection and propagation *up* to some local `trio` task nursery which is not the same nursery responsible for creating subproc management / spawn tasks

yes! this allows us to drop the iface from `ActorNursery` (and/or possibly `Portal`? though not sure on this yet) as needed - which would reduce surface area and the current complexity of `.run_in_actor()` being couple to multiple runtime layers.

> - a simpler API where we can demonstrate how to build the higher level constructs [like a worker pool](https://github.com/goodboy/tractor/blob/master/examples/parallelism/concurrent_actors_primes.py) from primitives and **there is less ways to do the same thing**.
 
> Also linking #172 as maybe another tidbit to consider as *sugar* level API. 

---

Also see nods in the doc regarding this,
https://goodboy.github.io/tractor/api/core.html#tractor.ActorNursery.run_in_actor

and the original deprecation PR,
https://github.com/goodboy/tractor/issues/290

goodboy:
### Motivation

The legacy non-blocking `ActorNursery.run_in_actor()` predates
streaming support and divided one-shot task ownership across several
runtime layers. It returned a `Portal` whose main result was owned by
hidden portal state and a nursery-exit reap task racing process
teardown inside each spawn backend.

Issue #477 prescribed moving one-shot invocation into the caller's
task. Merged PR #481 introduced that blocking successor as
`tractor.to_actor.run()`. This PR migrates every in-repository caller
and removes the old API plus its result/reaper machinery.

Attempting to retain the old API while simplifying supervision exposed
an unbounded `wait_for_result()` hang when a one-shot child was
cancelled out of band. Reintroducing another nursery-owned race would
preserve the wrong ownership model; caller-scoped result waiting and
cancellation are the structured-concurrency boundary.

### Source of research

- Issue #477 and the earlier design discussion:
  https://github.com/goodboy/tractor/issues/287#issuecomment-1005251485
- The implementation plan, machinery map, race probes and reap-hang
  post-mortem:
  `ai/conc-anal/ria_nursery_removal_plan.md`

### Summary of changes

- simplify actor-nursery supervision in three gated steps: move
  one-shot reaping out of spawn backends, remove the secondary
  `._ria_nursery`, and merge duplicate scope-error handlers.
- remove `ActorNursery.run_in_actor()`, portal main-result state,
  `Portal.wait_for_result()`, deprecated `Portal.result()`, ria reap
  helpers and the `NoResult` sentinel. The unrelated
  `Context.wait_for_result()` remains.
- migrate all in-repository tests, examples, debugging scripts,
  documentation and the experimental pubsub example to blocking
  `to_actor.run()`, daemon `start_actor()` plus `Portal.run()`, or
  linked contexts according to ownership.
- move fan-out into local Trio task nurseries. Their cancel-on-first
  semantics replace nursery-exit reap-all error collection, so raced
  siblings may be absorbed as cancellation rather than forming an
  exact BEG-of-N.
- retain high-fan-out startup/cancellation coverage with 25 concurrent
  callers and 1, 5 and 25 errorer variants. Verify bounded teardown,
  boxed assertion relays and empty actor-nursery child/reap maps.
- retain nested cancel-cascade coverage with race-tolerant error-shape
  validation and document the remaining MTF relay-versus-cancel race.
- fix the mutual-rendezvous lifetime race exposed by migration: peers
  which dial each other now use long-lived daemons, concurrent portal
  calls and explicit nursery cancellation.
- add anti-hang guards, stronger cancellation/result assertions,
  supervision diagrams, release notes and regression provenance.

### TODOs before landing

- [x] CI matrix green on the final tip (all legs: ubuntu
  tcp+uds, macos tcp, sdist, sphinx).
- [ ] recommend a `-x`-less full-suite pass on a normal-speed box if
  local runs looked thermally throttled (see the plan doc's B2
  notes).

### Future follow up

- [ ] worker-pool sugar (#172) and deferred-result parity belong in
  the `to_actor.open_taskman()` design tracked by #485.
- [ ] drop `test_nested_multierrors`' MTF shape-mismatch xfail once
  MTF runs consistently xpass.
- [ ] the pre-existing full-session `mp_spawn` failure belongs in the
  #379-era backend-revival work.

### Links

- resolves the remaining checklist in #477
- follows merged PR #481, which introduced `tractor.to_actor.run()`
- follow-up task-manager and worker-pool design: #485

(this pr content was generated in some part by
[`claude-code`][claude-code-gh] and revised by `opencode` using
`gpt-5.6-sol` (`openai`))


goodboy (commented):
### [P2] Pass the pub function directly to `open_stream_from()`

`tractor/experimental/_pubsub.py:228-234` | confidence: high | category: correctness

The example passes a `functools.partial` as the RPC target. `open_stream_from()` resolves it through `NamespacePath.from_ref()`, which identifies it as `functools:partial`, not `pub_service`. The remote actor does not expose `functools`, so the documented example cannot open its stream.

Recommendation: pass `pub_service` directly and provide `topics=` and `task_name=` as `open_stream_from()` keyword arguments.

### [P2] Reap the daemon in the pubsub example

`tractor/experimental/_pubsub.py:223-236` | confidence: high | category: reliability

The migration replaces `run_in_actor()` with `start_actor()`, creating a daemon. Closing `open_stream_from()` cancels only the stream task; the example never calls `portal.cancel_actor()`. If the consumer exits the stream loop, actor-nursery teardown waits indefinitely for the daemon.

Recommendation: cancel the actor in a `finally` block.

### [P2] Wait for pub/sub tasks to start before testing cancellation

`tests/test_advanced_streaming.py:235-283` | confidence: high | category: testing

The migrated test background-schedules every `to_actor.run()` call, sleeps three seconds, then raises the expected cancellation exception. It never confirms that the publisher and consumers started. A slow backend can still be spawning children when cancellation arrives, allowing the test to pass without exercising cancellation of active pub/sub streams.

Recommendation: add explicit remote readiness synchronization before starting the cancellation delay.

### [P2] Do not suppress unexpected subscriber failures

`tests/test_pubsub.py:202-216` | confidence: high | category: testing

`_run_subs()` suppresses every `RemoteActorError` and `ContextCancelled`, including failures before explicit teardown. In the `pub_actor='streamer'` case, a subscriber can fail during registration or streaming and the test can continue to cancellation and pass.

Recommendation: record when teardown begins and suppress only cancellation attributable to `cancel_actor()`; re-raise earlier failures.

### [P3] Keep active RPC tasks in the SIGINT regression

`tests/test_cancellation.py:824-838` | confidence: high | category: testing

The old test ran three `sleep_forever()` remote tasks. Its replacement creates three idle daemons, so it no longer verifies that SIGINT cancellation reaches active remote task trees. A regression that cancels idle runtimes but strands RPC tasks could pass.

Recommendation: background-schedule a sleeping linked task in every daemon and wait for startup before sending SIGINT.

### [P3] Preserve the multi-error assertion

`tests/test_cancellation.py:162-208` | confidence: high | category: testing

`test_multierror` now accepts any `BaseExceptionGroup` or `RemoteActorError` without checking member count or boxed type. It passes if only one actor starts, if one actor raises an unrelated error, or if startup itself fails.

Recommendation: synchronize both erroring contexts and assert either the intended multiple `RemoteActorError[AssertionError]` members or a specifically validated cancel-on-first result.

### [P3] Add release notes for removed public methods

`docs/api/core.rst:58-67` | confidence: high | category: compatibility

The branch removes `ActorNursery.run_in_actor()`, `Portal.wait_for_result()`, and `Portal.result()`, but adds no `nooz/` fragment describing those breaking removals. The existing fragment only announces `to_actor.run()`.

Recommendation: add a fragment listing each removed API and its migration path.

**Assumptions**
- The hard API removal itself is intentional, so I did not flag the absence of compatibility shims.
- The `pub()` docstring example is intended to be runnable by users.

**Checks**
- Reviewed `74cbee6c...09e78ad0` statically, including lifecycle, migration, public API, docs, examples, and tests.
- Range comparison found no duplicated #481 commits or accidentally dropped #484 behavior.
- Ruff passes on changed Python paths except three pre-existing unused imports in `examples/integration/open_context_and_sleep.py`, whose branch change only renames identifiers inside comments.
- Tests were not run because this worktree lacks its own active pytest environment.

**Scope**
Local/prospective review at `09e78ad087b54b0864c58862f69b431b44e35572`; base and merge base `74cbee6c697287738fa60f17ce9810252a322a71`. Provenance files under `ai/` were excluded from behavioral review.

(this review was generated in some part by `OpenCode` using `gpt-5.6-sol` (`openai`))

ai/conc-anal/ria_nursery_removal_plan.md
```diff
@@ -0,0 +1,460 @@
...
+                            `test_trynamic_trio` flake).
+
+Semantic deltas (tests loosened accordingly):
+
+- teardown-reap-all BEG-of-N is GONE: local task-nurseries are
+  cancel-on-first, raced siblings' `Cancelled`s are absorbed,
+  and the runtime's `collapse_eg()` unwraps every single-member
+  group at each actor boundary — a fully-raced nested tree
+  relays a bare (annotated) `RemoteActorError` chain.
+- `test_multierror_fast_nursery` deleted (pure reap-stress);
```
goodboy: ahh i see it was removed? so we're assuming nested_multierrors covers the original case entirely then ya?

tests/test_cancellation.py
```diff
@@ -104,56 +126,37 @@ def test_remote_error(
...
+        # directly in this caller task.
+        with pytest.raises(
             tractor.RemoteActorError,
-        )) as excinfo:
+        ) as excinfo:
             trio.run(main)
 
-        # ensure boxed errors are `errtype`
-        err: BaseException = excinfo.value
-        if isinstance(err, BaseExceptionGroup):
```
goodboy, line 150: so we don't need to worry about processing eg's in this test any more?

tests/test_cancellation.py
```diff
@@ -162,113 +165,48 @@ def test_multierror(
...
-    'delay',
-    (0, 0.5),
-    ids='delays={}'.format,
-)
-@pytest.mark.parametrize(
-    'num_subactors',
-    range(25, 26),
-    ids= 'num_subs={}'.format,
-)
-def test_multierror_fast_nursery(
```
goodboy, line 202: wait, what happened to this suite?

tests/test_cancellation.py
```diff
@@ -418,51 +366,63 @@ async def test_some_cancels_all(
...
+                            # can be scheduled)
+                            num_actors = 1
+                            # reraise so nursery teardown is triggered
+                            raise
+                        else:
+                            if expect_error:
+                                pytest.fail(
+                                    "Deamon call should fail at checkpoint?")
+
+        # should error here with a `RemoteActorError` or a beg of them
```
goodboy: maybe de-dent this and move directly above the next handler line?

tests/test_cancellation.py
```diff
@@ -475,19 +435,33 @@ async def spawn_and_error(
     breadth: int,
     depth: int,
 ) -> None:
+    '''
+    Recursively spawn a breadth-wide level of erroring one-shot
```
goodboy, line 617: can we maybe toss in a little ascii diagram of the supervision tree and have it indicate the remote-err propagation somehow?

tests/test_cancellation.py
```diff
@@ -588,6 +573,13 @@ async def test_nested_multierrors(
     # fork-spawn jitter + UDS-contention widens both `t1` and
     # `t2` further.
     #
+    # NB post-#477 (`to_actor.run()` fan-out in a local
+    # task-nursery) a race-tripped sibling's `Cancelled` is
+    # ABSORBED by the task-nursery instead of landing in the
+    # group — the raced case now shows as a *smaller* BEG, so
+    # this marker should consistently `xpass`; drop it once CI
```
goodboy, line 758: ahh so we haven't resolved this race already?

tests/test_pubsub.py
```diff
@@ -176,33 +176,55 @@ def test_multi_actor_subs_arbiter_pub(
...
+                    await portal.run(
+                        subs,
+                        which=which,
+                        pub_actor_name=name,
+                    )
+                except (
+                    tractor.RemoteActorError,
+                    tractor.ContextCancelled,
+                ):
+                    pass  # expected once we `cancel_actor()` below
```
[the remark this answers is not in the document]
goodboy: > response authored by `OpenCode`

Fixed by giving each subscriber runner a per-portal teardown event and suppressing only `ContextCancelled` from the root actor after that portal's explicit cancellation starts. Generic `RemoteActorError`, foreign cancellation, and premature cancellation now propagate.

> 📎 fixed in [`1d59f196`](https://github.com/goodboy/tractor/commit/1d59f1963c61013e412e9a46c8754f8fef9078cf)

tests/test_cancellation.py
```diff
@@ -418,51 +366,63 @@ async def test_some_cancels_all(
...
+                    for portal in dactor_portals:
+                        # if this function fails then we should error
+                        # here and the nursery should teardown all
+                        # other actors
+                        try:
+                            await portal.run(func, **kwargs)
+
+                        except tractor.RemoteActorError as err:
+                            assert err.boxed_type == err_type
+                            # we only expect this first error to propogate
```
goodboy: wait did we not fix this yet? XD

tests/test_cancellation.py
```diff
@@ -418,51 +366,63 @@ async def test_some_cancels_all(
...
+                            # (all other daemons are cancelled before they
+                            # can be scheduled)
+                            num_actors = 1
+                            # reraise so nursery teardown is triggered
+                            raise
+                        else:
+                            if expect_error:
+                                pytest.fail(
+                                    "Deamon call should fail at checkpoint?")
```
goodboy: nor this?

tests/test_infected_asyncio.py
```diff
@@ -219,18 +246,34 @@ def test_tractor_cancels_aio(
...
-            )
-            # cancel the entire remote runtime
-            await portal.cancel_actor()
+        # anti-hang wall-clock cap: a per-test `trio.fail_after`
+        # is the blessed guard here since `pytest-timeout`'s
+        # global cap is intentionally off (see the `pyproject`
+        # NOTE — it breaks trio under fork backends). Generous +
+        # CPU-headroom-scaled bc this is an anti-hang guard, not
+        # a perf assertion; a wedged ria-reaper once hung this
+        # test forever (the `._ria_nursery`-removal regression).
```
goodboy, line 255: can we link to the regression fix by chance somehow?

tests/test_spawning.py
```diff
@@ -48,38 +48,30 @@ async def spawn(
...
-                data=data_to_pass_down,
-                reg_addr=reg_addr,
             )
-
-            assert len(an._children) == 1
-            assert (
-                portal.channel.aid.uid
-                in
-                tractor.current_actor().ipc_server._peers
-            )
```
goodboy, line 71: hmm any way we can keep this internal check? if you think it's already covered well elsewhere and doesn't need repeating here, then that's fine too.

tests/test_spawning.py
```diff
@@ -159,21 +151,18 @@ async def test_most_beautiful_word(
...
-
-            res: Any = await portal.wait_for_result()
             assert res == return_value
-    # The ``async with`` will unblock here since the 'some_linguist'
-    # actor has completed its main task ``cellar_door``.
-
-    # this should pull the cached final result already captured during
-    # the nursery block exit.
-    res: Any = await portal.wait_for_result()
-    assert res == return_value
```
goodboy, line 176: no more checking the return val?

tractor/runtime/_portal.py
```diff
@@ -137,102 +127,6 @@ def channel(self) -> Channel:
         )
         return self.chan
 
-    # TODO: factor this out into a `.highlevel` API-wrapper that uses
-    # a single `.open_context()` call underneath.
-    async def _submit_for_result(
```
goodboy, line 142: ohh man feels sooo good to get rid of all this junk!

tests/test_pubsub.py
```diff
@@ -257,6 +279,9 @@ async def main():
             else:
                 await master_portal.cancel_actor()
 
+            # drop the bg `subs()` runners now the subs are cancelled
+            tn.cancel_scope.cancel()
```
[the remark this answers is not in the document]
goodboy: > response authored by `OpenCode`

Leaving this in place. Cancelling a nursery's own cancel scope from its host task is Trio's standard structured-shutdown pattern: cancellation is bounded and suppressed when that nursery scope exits. This is also the final statement in the inner nursery, so no subsequent host work can be skipped. Keeping it guarantees that any still-blocked `Portal.run()` runner is terminated even if actor cancellation does not make it return promptly.

ai/conc-anal/ria_nursery_removal_plan.md
goodboy: > response authored by `OpenCode`

Good catch: `test_nested_multierrors()` does not cover the original flat, high-fanout startup/cancellation load. I restored that useful part as `test_concurrent_start_error_reaps_all()`, with 25 concurrent `to_actor.run()` callers and 1/5/25 errorer variants. The obsolete requirement for an exact BEG-of-25 remains removed because caller-scoped cancel-on-first semantics cannot promise that shape.

> 📎 fixed in [`d5e66dff`](https://github.com/goodboy/tractor/commit/d5e66dffaa225071f8eb0d5df043114d10d7b7a5)

tests/test_cancellation.py
goodboy: > response authored by `OpenCode`

Right, not in this test anymore. Blocking `to_actor.run()` raises the child's boxed `RemoteActorError` directly in this caller task, while bad input arguments raise local `TypeError` before an actor is spawned. Exception-group shape and member validation remain covered by `test_multierror()`.

tests/test_cancellation.py
goodboy: > response authored by `OpenCode`

The old suite's exact BEG-of-25 assertion depended on the deleted nursery-exit reap-all mechanism, but deleting all of its stress coverage was too aggressive. I restored the startup/cancel/reap workload as `test_concurrent_start_error_reaps_all()`, validating bounded teardown, boxed assertion relays, and empty child/reap maps under the new caller-owned model.

> 📎 fixed in [`d5e66dff`](https://github.com/goodboy/tractor/commit/d5e66dffaa225071f8eb0d5df043114d10d7b7a5)

tests/test_cancellation.py
goodboy: > response authored by `OpenCode`

Yep, moved it out to align directly above the `except` handler.

> 📎 fixed in [`d5e66dff`](https://github.com/goodboy/tractor/commit/d5e66dffaa225071f8eb0d5df043114d10d7b7a5)

tests/test_cancellation.py
goodboy: > response authored by `OpenCode`

Added an ASCII supervision/error-flow diagram in the enclosing test documentation. It shows daemon and one-shot RAEs converging on the local Trio nursery, first-error sibling cancellation, and the escaping error causing `ActorNursery` to cancel and reap every child. The recursive helper keeps the depth-dependent grouping details in prose.

> 📎 fixed in [`d5e66dff`](https://github.com/goodboy/tractor/commit/d5e66dffaa225071f8eb0d5df043114d10d7b7a5)

tests/test_cancellation.py
goodboy: > response authored by `OpenCode`

Not fully. #477 changes the observable shape because the local task nursery absorbs a raced sibling's `Cancelled`, so the result is generally a smaller BEG rather than `[RAE, Cancelled]`. It does not close the underlying MTF relay-versus-cancel timing race, so the non-strict marker stays until MTF runs consistently xpass. The current GitHub matrix is Trio-only and cannot establish that yet.

tests/test_cancellation.py
goodboy: > response authored by `OpenCode`

Fixed already; this is the same `propogate` typo as resolved Copilot thread `3540613268`.

> 📎 fixed in [`cf56f33b`](https://github.com/goodboy/tractor/commit/cf56f33be6586df191270a217c64e156b04e5b1b)

tests/test_cancellation.py
goodboy: > response authored by `OpenCode`

Also fixed already; this duplicates resolved Copilot thread `3540613283`.

> 📎 fixed in [`cf56f33b`](https://github.com/goodboy/tractor/commit/cf56f33be6586df191270a217c64e156b04e5b1b)

tests/test_infected_asyncio.py
goodboy: > response authored by `OpenCode`

Linked the original regression-guard commit and the detailed removal analysis directly from the test comment. The original hash is `d1fb4a1a`; its current rebased PR-history counterpart is `61ad5bd1`.

> 📎 fixed in [`b3e8ed18`](https://github.com/goodboy/tractor/commit/b3e8ed18d11b937459c7a4a994e4f8875cdd5994)

tests/test_spawning.py
goodboy: > response authored by `OpenCode`

I did not restore the exact `_children`/`ipc_server._peers` check because `to_actor.run()` does not return until the one-shot child has completed and been reaped, so there is no live child or peer to inspect at that point. Reconstructing the old transient portal exposure made this API-level test substantially harder to understand. `test_spawn_from_caller_nursery()` already verifies post-return child/reap cleanup; the recursive test now focuses on its original root/child RPC, registrar, and result contracts.

tests/test_spawning.py
goodboy: > response authored by `OpenCode`

We still check the return value once, immediately after `to_actor.run()` returns. The removed second check only re-read the same cached portal result after nursery exit; that cache and `Portal.wait_for_result()` are part of the API being deleted.

tractor/runtime/_portal.py
goodboy: > response authored by `OpenCode`

Agreed. The portal main-result cluster is gone and no follow-up is needed here.

PR #488, opened 2026-08-04
Wait for ctx exit in `maybe_open_context()`

goodboy:
### Motivation

The final user of a cached `@acm` could return from
`maybe_open_context()` before the resource finished `__aexit__()`.
That exposed stale external state after the user's ctx closed,
detached regular cleanup errors from the user that triggered
teardown, and let a same-key entrant race the previous resource
generation's exit.

Complete the cache lifecycle at the user boundary: the final user now
waits for resource exit, receives regular cleanup errors there, and
keeps queued entrants serialized until replacement can safely begin.

---

### Summary of changes

- Add a `_CtxExit` handshake between `_Cache.run_ctx()` and the final
  `maybe_open_context()` user, publishing exit completion and regular
  `Exception` failures while preserving `BaseException` propagation.

- Serialize user deregistration and final teardown under the per-key
  `trio.StrictFIFOLock`; cancellation-shield the wait and retain the
  lock while same-key entrants remain queued for a fresh cache
  generation.

- Cover successful and failing cleanup, final-user and
  service-nursery cancellation, cleanup-error masking, and
  deterministic re-entry during teardown in
  `tests.test_resource_cache`.

- Record the OpenCode prompt/output provenance under
  `ai.prompt-io.opencode`.

---

---

(this pr content was generated in some part by `opencode` using
`gpt-5.6-sol` (`openai`))

### DIFF
--- a/tractor/trionics/_mngrs.py
+++ b/tractor/trionics/_mngrs.py
@@ -255,7 +255,7 @@ class _Cache:
                 try:
                     await no_more_users.wait()
                 finally:
-                    value = cls.values.pop(ctx_key)
+                    cls.values.pop(ctx_key)
                     cls.resources.pop(ctx_key)
 
         except Exception as exc:
@@ -403,11 +403,6 @@ async def maybe_open_context(
             resources = _Cache.resources
             entry: tuple|None = resources.get(ctx_key)
             if entry:
-                (
-                    service_tn,
-                    ev,
-                    ctx_exit,
-                ) = entry
                 raise RuntimeError(
                     f'Caching resources ALREADY exist?!\n'
                     f'ctx_key={ctx_key!r}\n'
--- a/tractor/trionics/_mngrs.py
+++ b/tractor/trionics/_mngrs.py
@@ -198,6 +198,16 @@ async def gather_contexts(
 # Further potential examples of interest:
 # https://gist.github.com/njsmith/cf6fc0a97f53865f2c671659c88c1798#file-cache-py-L8
 
+class _CtxExit:
+    '''
+    Completion state for a cached context's shared exit.
+
+    '''
+    def __init__(self) -> None:
+        self.done = trio.Event()
+        self.error: Exception|None = None
+
+
 class _Cache:
     '''
     Globally (actor-processs scoped) cached, task access to
@@ -213,7 +223,11 @@ class _Cache:
     values: dict[Any,  Any] = {}
     resources: dict[
         Hashable,
-        tuple[trio.Nursery, trio.Event]
+        tuple[
+            trio.Nursery,
+            trio.Event,
+            _CtxExit,
+        ],
     ] = {}
     # nurseries: dict[int, trio.Nursery] = {}
     no_more_users: trio.Event|None = None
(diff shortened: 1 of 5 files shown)


tractor/trionics/_mngrs.py
```diff
@@ -223,18 +237,38 @@ async def run_ctx(
...
+                    _,
+                    no_more_users,
+                    _,
+                ) = cls.resources[ctx_key]
+                cls.values[ctx_key] = value
+                task_status.started(value)
+                try:
+                    await no_more_users.wait()
+                finally:
+                    value = cls.values.pop(ctx_key)
```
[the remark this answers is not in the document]
goodboy: > response authored by `opencode`

Removed both dead bindings: `cls.values.pop(ctx_key)` now discards
directly, and the stale-resource assertion no longer unpacks unused
tuple entries.

> 📎 fixed in [`55ec3dbf`](https://github.com/goodboy/tractor/commit/55ec3dbf51c5265b10ab26029bfc0698e2103b07)

tractor/trionics/_mngrs.py
```diff
@@ -284,6 +318,8 @@ async def maybe_open_context(
     # sentinel = object()
```
goodboy: mm can we maybe remove this masked line here?
Also i think i had in my notes to adopt `anyio`'s use of `typing_extensions.sentinel` as well per,

https://github.com/agronholm/anyio/pull/1190/changes

not sure if that's applicable here given we already have `_UnresolvedCtx`?

tractor/trionics/_mngrs.py
```diff
@@ -445,44 +490,55 @@ async def maybe_open_context(
...
+                    # If no consumers remain, keep entrants queued
+                    # until the cached context has completely exited.
+                    if _Cache.users[ctx_key] <= 0:
+                        log.debug(
+                            f'De-allocating @acm-func entry\n'
+                            f'ctx_key={ctx_key!r}\n'
+                            f'acm_func={acm_func!r}\n'
+                        )
+
+                        # XXX: if we're cancelled we the entry may
```
goodboy: typo extra "we" and missing `,`?

tractor/trionics/_mngrs.py
```diff
@@ -445,44 +490,55 @@ async def maybe_open_context(
...
+                                _Cache.locks.pop(ctx_key)
+                            else:
+                                log.error(
+                                    f'Resource lock for {ctx_key} '
+                                    f'was replaced before teardown?'
+                                )
+                finally:
+                    lock.release()
+
+        if exit_error is not None:
```
goodboy, line 542: maybe a comment briefly indicating that we always reraise any `.__aexit__()` error?

goodboy:
Aight, one more CI and i think this puppy is good to land 😼

tractor/trionics/_mngrs.py
goodboy: > response authored by `opencode`

Added a comment documenting that a stored regular `__aexit__()` error
is always re-raised at the final consumer's context boundary.

> 📎 fixed in [`daa661ab`](https://github.com/goodboy/tractor/commit/daa661aba3708901cd0268b45f2ac5b920146101)

tractor/trionics/_mngrs.py
goodboy: > response authored by `opencode`

Fixed the wording to read: “if we're cancelled, the entry may have
never been entered”.

> 📎 fixed in [`daa661ab`](https://github.com/goodboy/tractor/commit/daa661aba3708901cd0268b45f2ac5b920146101)

tractor/trionics/_mngrs.py
goodboy: > response authored by `opencode`

Removed the masked sentinel experiment. I kept `_UnresolvedCtx` as
the internal marker rather than adding a
`typing_extensions.Sentinel` runtime dependency; the existing
identity marker is sufficient here.

> 📎 fixed in [`daa661ab`](https://github.com/goodboy/tractor/commit/daa661aba3708901cd0268b45f2ac5b920146101)

